PR #9, opened 2016-02-01
use options.Pipeline instead of just build

termie:


mwhooker:
:+1:

faiq:
:+1: tests pass locally although im not quite sure what code is calling the docker/builder.go to create the services

termie:
kk, merging this, i don't see any more TODOs from my packaging refactor that make me feel like this won't work

PR #11, opened 2016-02-02
Update locations of variables

hatchan:
Also add `=` to solve warning.


hatchan:
Also, I'm not sure if we want to continue with the auto versioning based on day... Kept it in there just in case.

termie:
lgtm

PR #37, opened 2016-02-03
replace strings that have hidden values with a blank before printing

faiq:
Not really sure this is how you wanted me to do this but sanitizes all output for hidden vars (i assume the log gets called after everything has already been interpolated)


tests/projects/print-protected/ENVIRONMENT
```diff
@@ -0,0 +1 @@
+XXX_SECRET="this is really secret"
```
termie: maybe add an X_NOTSECRET and test that it gets shown

tests/projects/print-protected/wercker.yml
```diff
@@ -0,0 +1,11 @@
+build:
+    box:
+        id: busybox
+        cmd: /bin/sh
+    steps:
+        - script:
+            code: |
+                echo "$SECRET"
```
termie: please echo it in another string

event/literalloghandler.go
```diff
@@ -56,10 +58,23 @@ func (h *LiteralLogHandler) Logs(args *core.LogsArgs) {
...
-		}).Printf("%s %6s %q", shown, args.Stream, args.Logs)
+		}).Printf("%s %6s %q", shown, args.Stream, h.sanitizeLogs(args).Logs)
 	} else if h.shouldPrintLog(args) {
-		h.l.Print(args.Logs)
+		h.l.Print(h.sanitizeLogs(args).Logs)
+	}
+}
+
+// goes through the
+func (h *LiteralLogHandler) sanitizeLogs(args *core.LogsArgs) *core.LogsArgs {
```
termie: the logs go to more places than this, you probably want to do this sanitization in core/events.go#238

tests/projects/print-protected/wercker.yml
termie: you can probably grep the output log for your secret and make sure it didn't show up

event/literalloghandler.go
```diff
@@ -56,10 +58,23 @@ func (h *LiteralLogHandler) Logs(args *core.LogsArgs) {
...
 			"Stream": args.Stream,
-		}).Printf("%s %6s %q", shown, args.Stream, args.Logs)
+		}).Printf("%s %6s %q", shown, args.Stream, h.sanitizeLogs(args).Logs)
 	} else if h.shouldPrintLog(args) {
-		h.l.Print(args.Logs)
+		h.l.Print(h.sanitizeLogs(args).Logs)
+	}
+}
+
+// goes through the
```
hatchan: The commit is cut off

hatchan:
@faiq the log are already interpolated.

It might be a good idea to replace the empty string with something like `[hidden from logs]`, to indicate that there is a value (just not visible in logs). For users that are trying to debug an issue, and who might try to do a `echo $SECRET` it will be empty. They might think that the env var is not set. (which remind me, it might be a good idea to print the values during the setup environment step but not show the values)

I do have some reservation about this feature: 

I don't think this will work with more complex values, such as values with references to other env vars, such as `SECRET=$ANOTHER_SECRET`. This value will be interpolated in the shell, and thus give another value in the logs, so it won't be sanitized. Although this might seem logical that we do not sanitize this, users might expect that it will be sanitized.

Another complex value, ssh keys, contains shell escape stuff, which might not match the original value. Also these are contained over multiple lines, and I think that we're receiving logs line by line. Which mean we need to buffer n lines.

Given these pitfalls it might not be a good idea, to make the promise that hidden env vars won't show up in logs.

hatchan:
https://xkcd.com/1638/

termie:
technically in the SECRET=$ANOTHERSECRET case it'll still get sanitized because we know the value of the $ANOTHERSECRET

hatchan:
@termie not if it is defined inside container

termie:
@hatchan sure it will, the string we print is what gets sanitized, not what happens in the container

ANOTHERSECRET=foo
SECRET=$ANOTHERSECRET
# prints "foo" which will be sanitized
echo $SECRET

tests/projects/print-protected/wercker.yml
faiq: right, i did that here inside of the test file. 

 `$wercker --environment ${testDir}/ENVIRONMENT build $testDir > $logFile && grep -q "HIDDEN" $logFile`

faiq:
Added tests to address multiline stuff and defining things in the container, seems to work fine

PR #39, opened 2016-02-03
fix deploy targets, add regression tests

termie:


termie:
once this finishes building on master i'll push to prod

PR #44, opened 2016-02-03
Add aftersteps to the correct slice

### LINKED ISSUE #43: wercker does not report steps in the aftersteps section as aftersteps
The step are being reported but instead of being part of the afterSteps they are part of the main step which causes the issues described in #42 #41.

The old legacy stack does not have the same issue.

hatchan:
Resolves #43


flenter:
lgtm

Tonnu:
:+1:

PR #45, opened 2016-02-03
Fix scratch n push

hatchan:
Also fixes some issues with `--docker-local`

### DIFF
--- a/docker/docker.go
+++ b/docker/docker.go
@@ -664,21 +664,23 @@ func (s *DockerScratchPushStep) Execute(ctx context.Context, sess *core.Session)
 		ServerAddress: s.authServer,
 	}
 
-	checkOpts := CheckAccessOptions{
-		Auth:       auth,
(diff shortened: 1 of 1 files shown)


docker/docker.go
```diff
@@ -1019,39 +1022,37 @@ func (s *DockerPushStep) Execute(ctx context.Context, sess *core.Session) (int,
 }
 
 func (s *DockerPushStep) tagAndPush(imageID string, e *core.NormalizedEmitter, client *DockerClient, auth docker.AuthConfiguration) (int, error) {
-	if !s.dockerOptions.DockerLocal {
```
hatchan: Just removed the check, which caused everything else to collapse (there shouldn't be any changes in the code below)

Tonnu:
lgtm :+1:

hatchan:
Tested it locally, and it worked

flenter:
lgtm

PR #48, opened 2016-02-03
make sure the step order is incremented

termie:
this hasn't changed in forever, we ran into this while trying to debug a
reporting error for after steps after a failed build, but this number
should be updated anyway


faiq:
:+1:

mwhooker:
oh ho! :+1:

hatchan:
:100:

PR #49, opened 2016-02-04
Allow local users to attach Volumes to their wercker containers

faiq:
experiment to see if docker runs in docker

### DIFF
--- a/docker/box.go
+++ b/docker/box.go
@@ -97,16 +97,6 @@ func NewDockerBox(boxConfig *core.BoxConfig, options *core.PipelineOptions, dock
 	if err != nil {
 		return nil, err
 	}
-	var interpolatedVols []string
-	if options.EnableVolumes {
-		vols := util.SplitSpaceOrComma(boxConfig.Volumes)
-		env := options.HostEnv
-		interpolatedVols = make([]string, len(vols))
-		for i, vol := range vols {
-			interpolatedVols[i] = env.Interpolate(vol)
-		}
-	}
-
 	return &DockerBox{
 		Name:            name,
 		ShortName:       shortName,
@@ -120,7 +110,7 @@ func NewDockerBox(boxConfig *core.BoxConfig, options *core.PipelineOptions, dock
 		logger:          logger,
 		cmd:             cmd,
 		entrypoint:      entrypoint,
-		volumes:         interpolatedVols,
+		volumes:         []string{},
 	}, nil
 }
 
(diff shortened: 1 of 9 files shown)


termie:
good PoC

sgoings:
Some thoughts from the peeps @ Docker: https://github.com/jpetazzo/dind

docker/box.go
```diff
@@ -168,6 +170,9 @@ func (b *DockerBox) binds() ([]string, error) {
 			// volumes[b.options.MntPath(entry.Name())] = struct{}{}
 		}
 	}
+	for _, volume := range b.volumes {
```
termie: keep in mind this has to be behind a check for options.EnableDevSteps, it should only be enable for dev mode

termie: it can be it's own flag, enable-volumes if we want

docker/shellstep.go
```diff
@@ -130,5 +131,5 @@ func (s *ShellStep) ReportPath(...string) string {
 
 // ShouldSyncEnv before running this step = TRUE
 func (s *ShellStep) ShouldSyncEnv() bool {
-	return true
+	return false
```
termie: this can be true again, you'll want to rebase on master

docker/shellstep.go
```diff
@@ -59,10 +59,11 @@ func NewShellStep(stepConfig *core.StepConfig, options *core.PipelineOptions, do
 	})
 
 	return &ShellStep{
-		BaseStep: baseStep,
-		options:  options,
-		data:     stepConfig.Data,
-		logger:   util.RootLogger().WithField("Logger", "ShellStep"),
+		BaseStep:      baseStep,
+		options:       options,
+		dockerOptions: dockerOptions,
```
termie: once you rebase on master this should disappear

test-all.sh
```diff
@@ -51,7 +51,7 @@ testDirectMount() {
 
 runTests() {
   basicTest "source-path" build $testsDir/source-path || return 1
-  basicTest "test local services" build $testsDir/local-service/service-consumer || return 1
+  basicTest "test local services" --debug build  $testsDir/local-service/service-consumer || return 1
```
termie: we might consider just having debug on all the time for these

wercker.yml
```diff
@@ -1,4 +1,6 @@
-box: tcnksm/gox:1.5.1
+box:
+  id: tcnksm/gox:1.5.1
+  volumes: ["/var/run/docker.sock", "/Users/faiq/go/src/github.com/wercker"]
```
termie: the users faiq section probably needs to be passed in as an env var, which means these values will need to be interpolated, too

faiq: it does- which means that core/config is going to need some new code.

termie: also, this still is somewhat unlikely to work on the server where the working directories are randomized, so an env var will have to be passed in with the working dir  if we want to special case ourselves for testing

PR #51, opened 2016-02-04
test after step behavior

termie:


termie:
just adds tests so mergerating

PR #57, opened 2016-02-06
Faiq add scratch push

### LINKED ISSUE #50: write regression test for scratch n push
Make sure we test for the fixes in #45

faiq:
adds tests for scratch push by creating a dummy container with special tags and grepping for those tags from docker images

fixes #50

### DIFF
--- a/test-all.sh
+++ b/test-all.sh
@@ -63,6 +63,24 @@ testDirectMount() {
   fi
 }
 
+testScratchPush () {
+  echo -n "testing scratchnPush.."
+  testDir=$testsDir/scratch-n-push
+  logFile="${workingDir}/scratch-n-push.log"
+  grepString="uniqueTagFromTest"
+  docker images | grep $grepString | awk '{print $3}' | xargs -n1 docker rmi -f > /dev/null 2>&1
+  $wercker build $testDir --docker-local --working-dir $workingDir > $logFile && docker images | grep -q $grepString
+  if [ $? -eq 0 ]; then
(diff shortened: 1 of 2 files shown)


test-all.sh
```diff
@@ -63,6 +63,24 @@ testDirectMount() {
   fi
 }
 
+testScratchPush () {
+  echo -n "testing scratchnPush.."
+  testDir=$testsDir/scratch-n-push
```
hatchan: The `testDirectMount` above uses all lower case, this one still uses camel case.

tests/projects/scratch-n-push/wercker.yml
```diff
@@ -0,0 +1,8 @@
+build:
+  box:
+    id: busybox
+    cmd: /bin/sh
+  steps:
+    - internal/docker-push:
+        tag: uniqueTagFromTest, alsouniqueTagFromTest
```
hatchan: If we're doing two tags, then we should also have 2 tests

hatchan:
:+1:

tests/projects/scratch-n-push/wercker.yml
faiq: by two tests do you mean two grep tests?

test-all.sh
faiq: whoops must have been a vim mistake! `testDirectMount` should be camelCase

termie:
looks bueno :)

PR #61, opened 2016-02-09
Allow users to mount local volumes

faiq:
allows users to mount local volumes through a `Volumes` section in their wercker.yml file

in order to run this, users must use a `--enable-volumes` flag

we leverage this feature to test wercker inside of a wercker build container as well

### DIFF
--- a/docker/box.go
+++ b/docker/box.go
@@ -177,11 +177,11 @@ func (b *DockerBox) binds(env *util.Environment) ([]string, error) {
 			interpolatedVols = append(interpolatedVols, env.Interpolate(vol))
 		}
 		b.volumes = interpolatedVols
+		for _, volume := range b.volumes {
+			binds = append(binds, fmt.Sprintf("%s:%s:rw", volume, volume))
+		}
 	}
 
-	for _, volume := range b.volumes {
-		binds = append(binds, fmt.Sprintf("%s:%s:rw", volume, volume))
-	}
 	return binds, nil
 }
 
--- a/docker/box.go
+++ b/docker/box.go
@@ -50,6 +50,7 @@ type DockerBox struct {
 	logger          *util.LogEntry
 	entrypoint      string
 	image           *docker.Image
+	volumes         []string
 }
 
 // NewDockerBox from a name and other references
@@ -96,7 +97,6 @@ func NewDockerBox(boxConfig *core.BoxConfig, options *core.PipelineOptions, dock
 	if err != nil {
 		return nil, err
 	}
-
 	return &DockerBox{
 		Name:            name,
 		ShortName:       shortName,
@@ -110,6 +110,7 @@ func NewDockerBox(boxConfig *core.BoxConfig, options *core.PipelineOptions, dock
 		logger:          logger,
 		cmd:             cmd,
 		entrypoint:      entrypoint,
+		volumes:         []string{},
 	}, nil
 }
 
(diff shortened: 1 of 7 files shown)


faiq:
for original POC see #49

test-all.sh
```diff
@@ -66,10 +66,9 @@ testDirectMount() {
 
 runTests() {
   basicTest "source-path" build $testsDir/source-path || return 1
-  basicTest "test local services" build $testsDir/local-service/service-consumer || return 1
+  basicTest "test local services" --debug build  $testsDir/local-service/service-consumer || return 1
```
termie: this one doesn't need debug when it is default now

wercker.yml
```diff
@@ -1,4 +1,6 @@
-box: tcnksm/gox:1.5.1
+box:
+  id: tcnksm/gox:1.5.1
+  volumes: $CODE_PATH /var/run/docker.sock
```
termie: please make this box only for the test pipeline (copy this to there and then remove the volumes from this one)

termie:
looks good, small changes before merge

termie:
.. and rebase probably

docker/box.go
```diff
@@ -168,6 +169,19 @@ func (b *DockerBox) binds() ([]string, error) {
...
+	if b.options.EnableVolumes {
+		vols := util.SplitSpaceOrComma(b.config.Volumes)
+		var interpolatedVols []string
+		for _, vol := range vols {
+			interpolatedVols = append(interpolatedVols, env.Interpolate(vol))
+		}
+		b.volumes = interpolatedVols
+	}
+
+	for _, volume := range b.volumes {
```
hatchan: We can move this inside the if statement, right?

test-all.sh
```diff
@@ -84,10 +84,9 @@ testScratchPush () {
 
 runTests() {
   basicTest "source-path" build $testsDir/source-path || return 1
-  basicTest "test local services" build $testsDir/local-service/service-consumer || return 1
+  basicTest "test local services" build  $testsDir/local-service/service-consumer || return 1
```
hatchan: The extra space is not needed

hatchan:
Is is possible to mount certain container paths to other paths on the host? Something like:

```
/some_path:/other_path
```

hatchan:
Is it also possible to use the volumes for services? That would be really neat, as that would make it possible to have persistent databases between tests.

hatchan:
Minor comments, looks to me :+1:

docker/box.go
faiq: yea we can.

faiq:
> Is is possible to mount certain container paths to other paths on the host? Something like:

> /some_path:/other_path

as of right now, its not. we need to use maps in wercker.yml in order to support this, its not a priority for me right now but its not a hard fix i think. 


> Is it also possible to use the volumes for services? That would be really neat, as that would make it possible to have persistent databases between tests.

I think so. mostly because services are based on top of boxes (assuming you have the `--enable-volumes` flag on you should be able to do it. but not too sure. tapping in @termie for more words.

anyways fixed the changes and will merge.

hatchan:
>> Is is possible to mount certain container paths to other paths on the host? Something like:
>> 
>> /some_path:/other_path

> as of right now, its not. we need to use maps in wercker.yml in order to support this, its not a priority for me right now but its not a hard fix i think.

Not necessary, we can just use the same notation that I've used above. Which I believe is the same as the one that Docker uses. 

```
volumes: /home/ /host_path/:/container_path/
```
In this example `/home/` from the host will be in the container on path `/home/`. And `/host_path/` will be mounted on `/container_path/`.

Tonnu:
@faiq did you write docs for this by any chance?

faiq:
@Tonnu nope net yet :/

PR #65, opened 2016-02-10
move working dirs to .wercker

termie:


termie:
addresses #8

termie:
actually i have a couple things i want to do to this before merge

termie:
lookin ready now

cmd/runner.go
```diff
@@ -196,6 +205,11 @@ func (p *Runner) EnsureCode() (string, error) {
 				if util.ContainsString(ignoreFiles, abspath) {
 					ignores = append(ignores, file.Name())
 				}
+
+				// TODO(termie): remove this warning after a while
+				if util.ContainsString(deprecatedPaths, abspath) {
```
faiq: I'm under the opinion that you should keep all the old ignore files in ignoreFiles, and still have this warning.

hatchan:
:lgtm:

cmd/runner.go
termie: if they old ignore files are in there, the user has no way to deal with having their own _builds directory, previously they could change working-dir to not conflict but with this change that would no longer work

Issue #16, opened 2016-02-02
Sentcli should not expand relative symlinks


faiq:
_From @hatchan on January 12, 2016 9:29_

Currently when a repository contains a relative symlink, it will be expanded to an absolute symlink _prior_ to moving it to the correct path. So the path that is contained in the absolute symlink doesn't exists there anymore.

This problem occurs both using the local cli, and on the hosted platform.

How to reproduce:

```
# Create an empty directory, cd into it, then:
mkdir some_dir
cd some_dir
ln -s ../README.md SYMLINK_TO_README.md
cd ..
echo "some contents" > README.md
```

Running `ls` and `cat` locally results in the following: 

```
$ ls -la some_dir
total 8
drwxrwxr-x 2 bvdberg bvdberg 4096 jan 12 10:08 .
drwxrwxr-x 8 bvdberg bvdberg 4096 jan 12 10:23 ..
lrwxrwxrwx 1 bvdberg bvdberg   12 jan 12 10:08 SYMLINK_TO_README.md -> ../README.md
$ cat some_dir/SYMLINK_TO_README.md
some contents
```

Use the following wercker.yml:

```
box: ubuntu
build:
  steps:
    - script:
        name: ls -la some_dir
        code: ls -la some_dir
    - script:
        name: cat some_dir/SYMLINK_TO_README.md
        code: cat some_dir/SYMLINK_TO_README.md
```

This results locally in ([debug log](https://gist.github.com/hatchan/caf473f2190494ce734f)):

```
$ wercker build
--> No Docker host specified, checking: unix:///var/run/docker.sock
--> Executing pipeline
--> Running step: setup environment
Pulling from library/ubuntu: latest
Digest: sha256:b53bb7b0d18842214ac7472c2a8801e8682c247d30f1ba4bab0083a2e2e091ea
Status: Image is up to date for ubuntu:latest
--> Running step: wercker-init
--> Running step: ls -la some_dir
total 12
drwxr-xr-x 2 root root 4096 Jan 12 09:23 .
drwxr-xr-x 4 root root 4096 Jan 12 09:23 ..
lrwxrwxrwx 1 root root   81 Jan 12 09:23 SYMLINK_TO_README.md -> /home/bvdberg/dev/hatchan/pull-request-test-project/some_dir/SYMLINK_TO_README.md
--> Running step: cat some_dir/SYMLINK_TO_README.md
cat: some_dir/SYMLINK_TO_README.md: No such file or directory
[...]

PR #67, opened 2016-02-12
change symlink behavior

termie:


hatchan:
This resolves #16, right? Either way, :lgtm:

mbrevda:
This works brilliantly, thanks!

PR #68, opened 2016-02-12
output artifacts locally with --artifacts

termie:


termie:
i'm not totally sure on the store s3 behavior yet so need to do a little more stuffness

termie:
yup s3 store behavior worked as expected

faiq:
:+1:

docker/artifact.go
```diff
@@ -38,13 +38,18 @@ type Artificer struct {
 func NewArtificer(options *core.PipelineOptions, dockerOptions *DockerOptions) *Artificer {
 	logger := util.RootLogger().WithField("Logger", "Artificer")
 
-	s3store := core.NewS3Store(options.AWSOptions)
+	var store core.Store
+	if options.ShouldStoreS3 {
+		store = core.NewS3Store(options.AWSOptions)
+	} else {
+		store = core.NewLocalStore(options.BuildPath("artifacts"))
```
faiq: where is this method defined?

termie:
updated per @faiq's suggestions

Issue #53, opened 2016-02-05
Windows support


umcodemonkey:
Docker and go both support Windows, there's no reason for the wercker-cli to not support it as well. But somewhere there's a dependency on `github.com/coreos/go-systemd/journal` that fails to build with

> vendor\github.com\coreos\go-systemd\journal\send.go:91: undefined: syscall.UnixRights

termie:
@umcodemonkey we send a variety of logs directly to journald when running on the servers, i'm not sure on an easy way to make the feature only optionally compiled in, but it does seem possible for you to get a proof of concept working by stripping out references to journald (we don't have easy access to windows computers to test on and are almost certainly making other linux-specific assumptions)

termie:
@umcodemonkey to be clear, it isn't currently a priority for our team, but the diff may not be too huge and i'm willing to take a look at it if you are interested in taking a swing at it

umcodemonkey:
Understood. I'll see what I can do.

whitecolor:
What about windows support?

xenoterracide:
I'd just be happy if it worked in `docker-machine` with directories mounted, but I see this

```
docker@default:/c/Users/xeno/IdeaProjects/platform$ wercker --debug build
INFO[0000]   docker| --> No Docker host specified, checking: /var/run/docker.sock
DEBU[0000]   events| BuildStarted
DEBU[0000]  options|   AWSOptions *core.AWSOptions
DEBU[0000]  options|     AWSAccessKeyID string =
DEBU[0000]  options|     AWSRegion string = us-east-1
DEBU[0000]  options|     AWSSecretAccessKey string =
DEBU[0000]  options|     S3Bucket string = wercker-development
DEBU[0000]  options|     S3PartSize int64 = 104857600
DEBU[0000]  options|   ApplicationID string = platform
DEBU[0000]  options|   ApplicationName string = platform
DEBU[0000]  options|   ApplicationOwnerName string = docker
DEBU[0000]  options|   ApplicationStartedByName string = docker
DEBU[0000]  options|   AttachOnError bool = false
DEBU[0000]  options|   Checkpoint string =
DEBU[0000]  options|   CommandTimeout int = 1500000
DEBU[0000]  options|   DefaultsUsed core.PipelineDefaultsUsed = {true}
DEBU[0000]  options|   DeployTarget string =
DEBU[0000]  options|   DirectMount bool = false
DEBU[0000]  options|   EnableDevSteps bool = false
[...]

PR #73, opened 2016-02-12
Preliminary windows support

umcodemonkey:
This PR is the beginning of work to address #53. I'm not sure I have the glide.yaml totally correct, because though it properly does not checkout the dependencies, it then complains that i can't update the commits correctly. But it correctly installs everything else, and `go build` successfully creates a `wercker.exe`, which I've been able to run against my local docker-machine.

Main concerns here are code-duplication; I don't have the changes factored as well as could be to remove redundancy between windows and non-windows versions of things. It's a judgement call on your end if you'd like more effort there. Comments welcome, and I'll continue to add improvements to this PR.

### DIFF
--- a/cmd/devflags.go
+++ b/cmd/devflags.go
@@ -1,0 +1,31 @@
+//   Copyright 2016 Wercker Holding BV
+//
+//   Licensed under the Apache License, Version 2.0 (the "License");
+//   you may not use this file except in compliance with the License.
+//   You may obtain a copy of the License at
+//
+//       http://www.apache.org/licenses/LICENSE-2.0
+//
+//   Unless required by applicable law or agreed to in writing, software
+//   distributed under the License is distributed on an "AS IS" BASIS,
+//   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//   See the License for the specific language governing permissions and
+//   limitations under the License.
+
+// +build !windows
+
+package cmd
+
+import (
+	"github.com/codegangsta/cli"
+)
+
+// These flags affect our local execution environment
+var DevFlags = []cli.Flag{
+	cli.StringFlag{Name: "environment", Value: "ENVIRONMENT", Usage: "Specify additional environment variables in a file."},
+	cli.BoolFlag{Name: "verbose", Usage: "Print more information."},
+	cli.BoolFlag{Name: "no-colors", Usage: "Wercker output will not use colors (does not apply to step output)."},
+	cli.BoolFlag{Name: "debug", Usage: "Print additional debug information."},
+	cli.BoolFlag{Name: "journal", Usage: "Send logs to systemd-journald. Suppresses stdout logging."},
+}
+
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,4 @@
 vendor
-wercker
+wercker*
 .wercker
 .DS_STORE
--- a/cmd/app.go
+++ b/cmd/app.go
@@ -1,0 +1,73 @@
+//   Copyright 2016 Wercker Holding BV
+//
+//   Licensed under the Apache License, Version 2.0 (the "License");
+//   you may not use this file except in compliance with the License.
+//   You may obtain a copy of the License at
+//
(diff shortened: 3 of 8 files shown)


termie:
sweeeet, i'll poke around on de-duping suggestions and get back to you

cmd/devflags.go
```diff
@@ -0,0 +1,31 @@
...
+	"github.com/codegangsta/cli"
+)
+
+// These flags affect our local execution environment
+var DevFlags = []cli.Flag{
+	cli.StringFlag{Name: "environment", Value: "ENVIRONMENT", Usage: "Specify additional environment variables in a file."},
+	cli.BoolFlag{Name: "verbose", Usage: "Print more information."},
+	cli.BoolFlag{Name: "no-colors", Usage: "Wercker output will not use colors (does not apply to step output)."},
+	cli.BoolFlag{Name: "debug", Usage: "Print additional debug information."},
+	cli.BoolFlag{Name: "journal", Usage: "Send logs to systemd-journald. Suppresses stdout logging."},
```
termie: i think you can probably leave this in and just say "Not applicable to Windows" as long as the windows app, above, doesn't check the value

termie: (by leave it in i mean we can leave this where it was in flags)

termie:
i don't actually understand how this works, how does it know which GetApp to use? is there a magic go feature thatunderstands _windows?

umcodemonkey:
There are two go features. One understands `_windows` and only builds that file on windows. The other reads the `// +build !windows` build constraint in the other files, and skips those files on windows. They aren't magic, but are documented as a part of [the `build` package](https://golang.org/pkg/go/build/).

termie:
neeeaaatt

termie:
also i consider that magic, but cool magic

termie:
i'd like to merge this, could you make the changes I asked for? (leave the flags alone except to add "not applicable on windows")

umcodemonkey:
Yes, just haven't had the cycles to get back to it. I'll push an updated commit later today.

F21:
I just build the wercker cli by merging this PR on top of master.

I also have docker toolbox set up using these instructions: http://devcenter.wercker.com/cli/installation/index.html

I then followed: http://devcenter.wercker.com/quickstarts/building/golang.html using the git shell (to support `eval` etc).

It failed on `wercker dev --publish 5000`:
```
$ wercker dev --publish 5000
--> Executing pipeline
Pulling from library/golang: latest
Already exists: 7268d8f794c4
Already exists: a3ed95caeb02
Already exists: d9a49bc2b1b0
Already exists: b965864d2d45
Already exists: bad3f2daf720
Already exists: a3ed95caeb02
Already exists: a3ed95caeb02
Already exists: a3ed95caeb02
Already exists: 3c0458acec93
Already exists: a3ed95caeb02
Already exists: a3ed95caeb02
Already exists: 312d8dc0bff1
Already exists: a3ed95caeb02
Already exists: 9d76f8609ef9
Digest: sha256:0325944f1c19cacc2d039cc8933d7cd4ed4651466a0765a23a127c6866a660da
Status: Image is up to date for golang:latest
ERROR --> Step failed: setup environment 7.76s
Command cancelled due to error
ERROR Command cancelled due to error
WARNING Box container has already stopped.
FATAL Exiting.
```

I am using Windows 10 64-bit, Virtualbox 5.0.14 and the latest Windows build of docker toolbox.

F21:
After some debugging, I found that the context times out when trying to `SetupGuest()`. The strange thing is that I have tried ` --no-response-timeout "10" --command-timeout "25"` and confirmed with debug statements that the timeouts are set correctly, but the wercker process still exits within a few seconds.

umcodemonkey:
It's possible this is interacting oddly with the windows pty deficiencies, and that using `winpty` in front of docker commands will be needed, but I hadn't had a chance to fully vet that.

F21:
I have done a bit more tracing and it looks like the context is being cancelled here: https://github.com/wercker/wercker/blob/master/docker/session.go#L79

whitecolor:
So will it be?

PR #74, opened 2016-02-12
local deploys with smart defaults

termie:


termie:
merging this so that it can live in staging for a little while

Tonnu:
@termie any idea when you want to release this? Thinking of writing up a blogpost

PR #79, opened 2016-02-17
Rename test asset

hatchan:


termie:
:+1:

faiq:
i came up with that name with love and this is how you do me @hatchan?!!

faiq:
it was a great name!

PR #85, opened 2016-02-23
dont add a pipeline tag for docker images

### LINKED ISSUE #84: Don't auto-tag docker images with the build id during internal/docker-push
A user has requested a way to turn off auto-tagging images with the build ID, since it's too verbose.

cc @termie 

uservoice #3793

faiq:
- fixes #84

### DIFF
--- a/docker/docker.go
+++ b/docker/docker.go
@@ -1010,7 +1010,6 @@ func (s *DockerPushStep) Execute(ctx context.Context, sess *core.Session) (int, 
 		Author:     s.author,
 		Message:    s.message,
 		Run:        &config,
-		Tag:        s.options.PipelineID,
 	}
 
 	s.logger.Debugln("Commit container:", containerID)


mwhooker:
is this the right way to do this? Why did we do this originally?

Are those reasons no longer valid?

maybe this should be a flag

faiq:
Sorry, for not getting to this earlier. So the reason we decided to have an automatic tag placed on the image was because we wanted to avoid having collisions with previously used tags on the same image repository name. For example, suppose we have `faiq/busybox` that was already built and tagged with `latest`, if we build another image for the same repository with a different image id and try to reuse the `latest` tag we'll get an error. This would result in the image being built but not having any tag. To avoid this, we give it a unique tag. But since we're force tagging now, i doesn't make it a difference now.

When these conflicts happen, the tag fails and tells you to try

termie:
i'm for this change, but i want to know what happens on the different versions of docker because they recently changed the tagging behavior

termie:
@faiq when you get a change can you figure out how this affects different versions of docker?

faiq:
okay so i tested with docker 1.9.1 by running the following pipeline twice and i was able to build and use the tag "yo" again successfully. this happened because we set `--force` to true

```
push:
    steps:
        - script:
            code: |
              cp wercker.yml /test
        - internal/docker-scratch-push:
            username: $DOCKER_U
            password: $DOCKER_P
            repository: quay.io/wercker/test-repo
            registry: https://quay.io
            tag: yo
```



this should also work fine with all new versions of docker, because the expected tagging behavior is now the expected behavior we want, that is being able to move a previously used tag from one container to another in the same repository.

see here for commit/discussion regarding deprecation of `-f`:
https://github.com/docker/docker/pull/18350

anyways, seems safe to merge now.

PR #91, opened 2016-02-29
Generate SHA256 hashes for binaries

hatchan:
This is a requirement for automating our brew deployment automation

### DIFF
--- a/wercker.yml
+++ b/wercker.yml
@@ -69,6 +69,18 @@ build:
               -output "$WERCKER_OUTPUT_DIR/latest/{{.OS}}_{{.Arch}}/sentcli"
 
             cp -r $WERCKER_OUTPUT_DIR/latest/* $WERCKER_REPORT_ARTIFACTS_DIR
+
+    # generate SHA's for all binaries
+    - script:
+        name: generate SHAs
+        code: |
+            cd $WERCKER_OUTPUT_DIR
+            for f in $(find . -type f -name sentcli); do
+            cd $WERCKER_OUTPUT_DIR/$(dirname "$f")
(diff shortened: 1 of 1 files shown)


wercker.yml
```diff
@@ -69,6 +69,18 @@ build:
               -output "$WERCKER_OUTPUT_DIR/latest/{{.OS}}_{{.Arch}}/sentcli"
 
             cp -r $WERCKER_OUTPUT_DIR/latest/* $WERCKER_REPORT_ARTIFACTS_DIR
+
+    # generate SHA's for all binaries
+    - script:
+        name: generate SHAs
+        code: |
+            cd $WERCKER_OUTPUT_DIR
+            for f in $(find . -type f -name sentcli); do
```
faiq: is the name still sentcli?

faiq: ya it is

wercker.yml
```diff
@@ -69,6 +69,18 @@ build:
               -output "$WERCKER_OUTPUT_DIR/latest/{{.OS}}_{{.Arch}}/sentcli"
```
faiq: maybe we should change the name wercker?

wercker.yml
hatchan: Yeah that is a separate issue. But not touching that one yet, as it is used on multiple locations

faiq:
:+1:

wercker.yml
hatchan: Thats going to be a bigger PR, which covers multiple projects. For now we're just doing a `sed` on the SHA256SUM files.

hatchan:
Thanks for the review \o/

PR #93, opened 2016-03-01
respect docker local on push

faiq:
if the docker local flag is on, we don't push to a remote repository.

from our docs:

```
Docker local
--docker-local
This flag can be used with either wercker dev or wercker build. When using this flag, wercker will not attempt to connect to a remote repository when pulling or pushing images, but will instead look for container images locally, or when using the internal/docker-push, it will push images to the local repository.
```

### DIFF
--- a/docker/docker.go
+++ b/docker/docker.go
@@ -515,7 +515,7 @@ func (s *DockerScratchPushStep) Execute(ctx context.Context, sess *core.Session)
 	if err != nil {
 		return -1, err
 	}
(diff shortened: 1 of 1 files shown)


termie:
probably güd

hatchan:
Looks good, think we also need to add it to the scratch and push step?

faiq:
@hatchan the code is shared between the two steps https://github.com/wercker/wercker/blob/faiq-docker-local-push/docker/docker.go#L694

PR #96, opened 2016-03-02
allow images to be pulled by local service builds

termie:
not sure why this was here originally, maybe @mwhooker can clarify


mwhooker:
Not sure either. looks like it was added in this commit https://github.com/wercker/wercker/commit/a7e0e1e0d769a15a1fe1388a04e8f90f19f81b41

termie:
hmm, that wasn't really the patch that added it, some patch a while before moved it but... hard to find, i'm willing to merge the change since it is local-only

PR #97, opened 2016-03-02
local deploys are now smarter about wercker.yml

termie:
checks target first, then defaults to cwd


mwhooker:
logic looks good but I don't have the context for this change

termie:
@mwhooker context is that when specifying a deploy target manually (rather than using the "automatic" target of .wercker/latest/output) you'd always have to specify a --wercker-yml, this makes the wercker-yml locating make more sense

PR #99, opened 2016-03-03
add checkpointing and base-path

termie:
base-path lets us get around many of the issues that setup-go-workspace ran
into due to having the change the name of the working dir, it can be set as
a field on a pipeline

checkpointing adds the ability to set checkpoints at steps causing the
container to get locally committed and adds a flag "--checkpoint" that will
cause the build to resume from that commit

also changed is we now remove the "source" path if it already existed in
the container, allowing builds to be chained a bit more easily

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -965,7 +965,21 @@ func executePipeline(cmdCtx context.Context, options *core.PipelineOptions, dock
 	// stepCounter starts at 3, step 1 is "get code", step 2 is "setup
 	// environment".
 	stepCounter := &util.Counter{Current: 3}
+	checkpoint := false
 	for _, step := range pipeline.Steps() {
+		// we always want to run the wercker-init step to provide some functions
+		if !checkpoint && stepCounter.Current > 3 {
+			if options.EnableDevSteps && options.Checkpoint != "" {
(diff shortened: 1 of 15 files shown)


cmd/main.go
```diff
@@ -965,7 +965,26 @@ func executePipeline(cmdCtx context.Context, options *core.PipelineOptions, dock
...
 	// environment".
 	stepCounter := &util.Counter{Current: 3}
+	checkpoint := false
 	for _, step := range pipeline.Steps() {
+		// we always want to run the wercker-init step to provide some functions
+		if !checkpoint && stepCounter.Current > 3 {
+			if options.Checkpoint != "" {
+				if step.Checkpoint() != options.Checkpoint {
+					logger.Printf(f.Info("Skipping step", step.DisplayName()))
+					stepCounter.Increment()
```
hatchan: The stepcounter increment and `continue` can be moved next to the other `if` statement (and removed from there as well)

termie: gotcha, since i was checking for two opposites i just moved the stepcount and continue down to bottom, the skipping step to top and got rid of the first if

termie:
mergerated

PR #100, opened 2016-03-04
Make working-dir environment-configurable.

### LINKED ISSUE #98: Support reading command line options from the env.
I think it would be really nice if the CLI would support getting some (all?) of it's options from the envrionement and it seems the cli library that is used [supports it](https://github.com/codegangsta/cli#values-from-the-environment).

Good example would be to change the build working-dir without having to specify it all the time.

mwhooker:
resolves #98

### DIFF
--- a/cmd/flags.go
+++ b/cmd/flags.go
@@ -44,7 +44,7 @@ var (
 
 	// These flags control where we store local files
 	LocalPathFlags = []cli.Flag{
-		cli.StringFlag{Name: "working-dir", Value: "./.wercker", Usage: "Path where we store working files."},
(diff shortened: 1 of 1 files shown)


cmd/flags.go
```diff
@@ -103,7 +103,7 @@ var (
 
 	// These flags affect our local execution environment
 	DevFlags = []cli.Flag{
-		cli.StringFlag{Name: "environment", Value: "ENVIRONMENT", Usage: "Specify additional environment variables in a file."},
+		cli.StringFlag{Name: "environment", Value: "ENVIRONMENT", Usage: "Specify additional environment variables in a file.", EnvVar: "WERCKER_ENVIRONMENT_FILE"},
```
termie: ftr, the environment file specified here won't change the environment that is parsed to do the defaults for the other flags

termie:
lgtm, merging

PR #103, opened 2016-03-09
allow multiple services with same image

termie:


mwhooker:
:lgtm:

hatchan:
:+1:

Tonnu:
There's no docs for this by any chance? If not I'll write something up and then we could merge this?

PR #106, opened 2016-03-12
fix cache size limits on docker 1.9+

termie:


mwhooker:
feel like making this a readercloser would be the best thing to do, but this lgtm

termie:
TeeReader gives off a Reader so i'm just gonna go with this :|

PR #111, opened 2016-03-14
provide messages about failed artifact storage

### LINKED ISSUE #110: Builds fail without an error when artifacts > 1GB
We should provide feedback in the log/output on why the build is failing.

Ideally this is done in the logs as well as via the `message` (that's the error/success text displayed on the build overview page)

termie:
resolves #110


mwhooker:
:+1:

hatchan:
:+1:

PR #116, opened 2016-03-15
report the tar path for artifacts

### LINKED ISSUE #115: Artifact download missing .tar extension
When downloading artifacts, something has changed recently and the file extension ".tar" is no longer added to the downloaded file. This means I always need to add it manually before opening.

I am downloading by clicking on the "Download aritfact" button in the build step. Have tested in both Chrome and Safari. I get a file called "output" (no extension)

Further information: This is only the case for recent builds (not sure the exact date but in the last week). For older builds (even of the same project) it still downloads a file called "artifacts.tar". I get the same behavior across different projects.

mwhooker:
resolves #115


faiq:
LGTM

termie:
also tm it lg

PR #119, opened 2016-03-17
registry v2 and ECR support

faiq:
still needs work, but throwing this up there for some review


core/config.go
```diff
@@ -32,17 +32,22 @@ type RawBoxConfig struct {
...
+	Tag           string
+	Cmd           string
+	Env           map[string]string
+	Username      string
+	Password      string
+	Registry      string
+	Entrypoint    string
+	URL           string
+	Volumes       string
+	AwsRegistryID string `yaml:"aws-registry-id"`
```
termie: Not my personal favorite, but go style says these should be

AWSRegistryID

termie: relatedly, we should consider a more extensible solution to this, this is a bunch of aws specific stuff, maybe we can have a Credentials field that is a map[string]string ?

PR #121, opened 2016-03-21
bumps s3 dependency

faiq:
we need to bump our aws-sdk-go dependency in order to be able to use the [check access library] (https://github.com/wercker/docker-check-access)

relatively simple stuff has changed between the two versions, mostly regarding how you set up an uploader


faiq:
ran the step, and was able to push an artifact to s3 properly... so merging upon :+1:

termie:
lgtm

PR #122, opened 2016-03-22
bumps s3 dependency

faiq:
- bumps aws go dependency for ferengi


faiq:
same code as #121, ran this to see if the code built properly and it did

PR #126, opened 2016-03-23
WIP: Allow push and pull from ECR and docker v2 

faiq:


core/config.go
```diff
@@ -21,9 +21,10 @@ import (
 	"strconv"
 	"strings"
 
-	"gopkg.in/yaml.v2"
-
+	"github.com/wercker/docker-check-access"
+	"github.com/wercker/wercker/auth"
 	"github.com/wercker/wercker/util"
+	"gopkg.in/yaml.v2"
```
hatchan: My `goimports` keeps this import in its own section, any reason why it was moved down here?

docker/box.go
```diff
@@ -501,31 +502,11 @@ func (b *DockerBox) Fetch(ctx context.Context, env *util.Environment) (*docker.I
...
-		b.logger.Errorln("Error during check access")
-		return nil, err
-	}
-
-	if !check {
-		b.logger.Errorln("Not allowed to interact with this repository:", b.repository)
-		return nil, fmt.Errorf("Not allowed to interact with this repository: %s", b.repository)
+	authenticator := b.config.Auth.ToAuthenticator()
+	check, err := authenticator.CheckAccess(env.Interpolate(b.repository), auth.Pull)
+	if !check || err != nil {
```
hatchan: Think `err` should be handled differently. Maybe the endpoint was not responding, which currently results in a misleading error message.

docker/box.go
```diff
@@ -538,11 +519,14 @@ func (b *DockerBox) Fetch(ctx context.Context, env *util.Environment) (*docker.I
...
 		OutputStream:  w,
 		RawJSONStream: true,
-		Repository:    env.Interpolate(b.repository),
+		Repository:    authenticator.Repository(env.Interpolate(b.repository)),
 		Tag:           env.Interpolate(b.tag),
 	}
-
-	err = client.PullImage(options, auth)
+	authConfig := docker.AuthConfiguration{
+		Username: authenticator.Username(),
```
hatchan: shouldn't these also be put through `env.Interpolate`? Or is that already done?

docker/docker.go
```diff
@@ -649,40 +535,21 @@ func (s *DockerScratchPushStep) Execute(ctx context.Context, sess *core.Session)
...
-		}
-
-		check, err := client.CheckAccess(checkOpts)
-		if err != nil {
-			s.logger.Errorln("Error during check access", err)
-			return -1, err
-		}
-		if !check {
+		check, err := s.authenticator.CheckAccess(s.repository, auth.Push)
+		if !check || err != nil {
```
hatchan: Same here as above

docker/docker.go
```diff
@@ -960,31 +870,17 @@ func (s *DockerPushStep) Execute(ctx context.Context, sess *core.Session) (int,
...
```
hatchan: Same as above

docker/docker.go
```diff
@@ -1044,11 +935,17 @@ func (s *DockerPushStep) tagAndPush(imageID string, e *core.NormalizedEmitter, c
 	}
 	pushOpts := docker.PushImageOptions{
 		Name:          s.repository,
-		Registry:      s.registry,
 		OutputStream:  w,
 		RawJSONStream: true,
 	}
 	if !s.dockerOptions.DockerLocal {
+		s.username = s.authenticator.Username()
```
hatchan: `env.Interpolate`?

core/config.go
faiq: no clue about how this happened, but i'll fix it

PR #128, opened 2016-03-25
Emit config before parsing pipeline

hatchan:
And on invalid yaml

### DIFF
--- a/cmd/runner.go
+++ b/cmd/runner.go
@@ -453,6 +453,11 @@ func (p *Runner) SetupEnvironment(runnerCtx context.Context) (*RunnerShared, err
 
 	// Grab our config
 	rawConfig, stringConfig, err := p.GetConfig()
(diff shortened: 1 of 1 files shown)


Tonnu:
lgtm :+1:

cmd/runner.go
```diff
@@ -454,12 +454,23 @@ func (p *Runner) SetupEnvironment(runnerCtx context.Context) (*RunnerShared, err
 	// Grab our config
 	rawConfig, stringConfig, err := p.GetConfig()
 	if err != nil {
```
mwhooker: can this be simplified to one call to `Emit` before this line?

hatchan: Changed as suggested

hatchan:
Changed according to feedback, and added a check to see if `stringConfig` is not empty.

PR #133, opened 2016-03-30
Output a descriptive error message when setup environment fails

gpoul:
The problem occurred with a docker base image where mkdir /pipeline fails.

WIthout this patch the output is just that "something" failed with exit code 1. - especially annoying in the web interface when there is no easy way to enable debug.

With this patch the output is:

`Guest command failed with exit code 1: mkdir -p "/pipeline"
`


hatchan:
Sorry I accidentally clicked merge :-( 

Regarding the PR, looks good to me.

faiq:
@hatchan no big deal, was planning on merging today anyways! @gpoul thanks for the pull request!

PR #137, opened 2016-04-06
fix scratch push 1.10

faiq:


mwhooker:
same comments wrt go vet as in #138 

can you briefly outline what the problem is you are trying to fix?

wercker.yml
```diff
@@ -213,3 +213,11 @@ run-tests:
...
                 cd $CODE_PATH
                 ./test-all.sh
+
+scratch:
+  box:
+    id: alpine
+    cmd: /bin/sh
+  steps:
+    - internal/docker-scratch-push:
+        repository: quay.io/go/fuckurself
```
mwhooker: this is a public repository

docker/docker.go
```diff
@@ -895,6 +878,7 @@ func (s *DockerPushStep) Execute(ctx context.Context, sess *core.Session) (int,
 func (s *DockerPushStep) tagAndPush(imageID string, e *core.NormalizedEmitter, client *DockerClient) (int, error) {
 	// Create a pipe since we want a io.Reader but Docker expects a io.Writer
 	r, w := io.Pipe()
+	fmt.Println("here")
```
mwhooker: remove

faiq:
@mwhooker sure thing. Docker on v1.10 > changed all of their layerID's to be content addressable from the contents of the tarball for that particular layer - then they use this sum to build another json file that they hash to get the imageID.This breaks the current implementation of the scratch push step because we create layerID on random and try to use this imageID to tag and push what we just created.

What this pull request does is two things:

1. Compute the sha256 of the contents in layer.tar for that particular layer 
2. Use the answer from this to build a new json file, and hash it to get the appropriate imageID

for details about the new spec look here:https://github.com/docker/distribution/pull/1068

hatchan:
I have done the following:
- Rebased on `ferengi`
- Peg every docker dependency on the latest released version (according to the git tags)
- Remove `nat.Port`, replaced with a `docker.Port` -> `nat.Port` transform function
- Optimize by combining writing to disk and generating the digest (saves another read)

That said, I'm not comfortable merging this in at this point. I want to get to know more about the image format.

Tonnu:
I've verified this is working by pushing to ECR. 👍

mies:
:partyparrot:

Tonnu:
One thing I did notice just now is that only one tag got pushed (I specified two). Don't know if this is specific for the ECR + scratch 'n push step or not. Will do some further investigation.

faiq:
@hatchan thanks for the refactor- looking a little cleaner and more efficient now 

@Tonnu I've had some success with this working before, but it's also failed a couple of times. I'm thinking we need some tests for this before merging. We can likely copy tests from the docker/image package on github.

Tonnu:
@faiq already mentioned this, but since we're running a newer version of docker on the Wercker platform, the scratch 'n push step now seems broken. Any chance we can get this merged soon?

mies:
also the production build is failing. Can we get this merged in? /cc @faiq @termie @mwhooker

termie:
If this will work against 1.12 go ahead and merge it, make sure it targets ferengi too

mies:
@faiq can you merge this?

PR #138, opened 2016-04-09
bump glide

faiq:
glide has been failing with VCS errors and this bump acknowledges and fixes them


mwhooker:
why are we disabling go vet?

termie:
this is covered in #146 and #147

PR #141, opened 2016-04-11
fix go vet

termie:
i think our old usage was deprecated


faiq:
:+1:

hatchan:
For go 1.6 this needs to be `go vet ....` See wercker/reporter#33

PR #142, opened 2016-04-11
Ensure current image in repository is pushed

faiq:
push everytime you tag to ensure not ALL the tags for the repository get pushed to a registry

before, if you had a repository with multiple tags they would all get pushed up. this PR stops that


hatchan:
Good thing we caught this before going live :lgtm:

termie:
lgtm merging

PR #143, opened 2016-04-12
Remove github.com/docker/reference

Tonnu:
Ran into this when building from source. `docker/reference` is not actually a package that can be installed.


faiq:
ill take a look at this - i added in a lot of the new docker dependencies in my last 2 projects

Tonnu:
thanks @faiq. @hatchan mentioned I had an outdated glide version, so updated that and it worked afterwards. Though might be still worth checking out what's up seeing github.com/docker/reference doesn't exist

termie:
this is covered by #146 and #147

PR #144, opened 2016-04-13
updates glide lock

faiq:
OK although this might not seem like a critical fix, it is. Let me explain the problem: a while back i created [docker-check-access](https://github.com/wercker/docker-check-access/) in order to do authentication checking against registry providers to see if users were allowed to access local images. 

Things worked fine, but I recently noticed that pushing to docker hub was returning 400's when doing the authentication check. 

To fix this, I pinned the version of docker-reg-client in my library to the same one as we use in master

https://github.com/wercker/wercker/blob/master/glide.yaml#L49
https://github.com/wercker/docker-check-access/commit/a4b9bb5f7e8804a12a2f671441b6cccf7520fd9d

### DIFF
--- a/glide.lock
+++ b/glide.lock
@@ -1,8 +1,8 @@
 hash: 4245507251ba285a3528099dab3563f1bcec405716925dc410d5e632e79b8475
-updated: 2016-03-28T12:33:00.906950379-07:00
+updated: 2016-04-12T20:21:09.103542093-07:00
 imports:
 - name: github.com/aws/aws-sdk-go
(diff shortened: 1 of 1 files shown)


faiq:
i no longer receive 400s when pining the version

also if you're interested in the implementation of the authentication look here: https://github.com/wercker/docker-check-access/blob/master/docker-auth-v1.go

it is almost exactly identical to what we're doing in this function.

again - only difference is the version of docker reg client

mwhooker:
lgtm

PR #147, opened 2016-04-25
switch to govendor

termie:
glide has been giving us a variety of problems:
1. unreliable library selection between versions
2. a variety of git-related errors on early git versions
3. different invocations between versions
4. sometimes huge 100+MB checkouts

govendor gives us a better way to cache (though we aren't using it yet in this project), much smaller checkouts (19MB), and generally feels more mature and explicit in its design


mwhooker:
would you mind adding a sentence or two on what precipitated the change?

faiq:
Also, a lot of our code depends on pinning certain version of libraries for example docker-reg-client was pinned to support authz for pushing to docker hub. Can you make sure that tests pass properly?

PR #148, opened 2016-04-26
add docker constraints

termie:


docker/options.go
```diff
@@ -25,11 +25,18 @@ import (
...
-	DockerTLSVerify string
-	DockerCertPath  string
-	DockerDNS       []string
-	DockerLocal     bool
+	DockerHost         string
+	DockerTLSVerify    string
+	DockerCertPath     string
+	DockerDNS          []string
+	DockerLocal        bool
+	DockerCPUPeriod    int64
```
hatchan: Having everything prefixed with `Docker` seems a bit repetitive, since it already is `DockerOptions`. It could be removed from these new properties, but that means it won't be consistent with the previous values.

wercker.yml
```diff
@@ -1,23 +1,13 @@
...
-        code: |-
-          echo "welcome to your shell"
-          echo "stay a while, and listen"
+  id: golang
+
 build:
   # base-path: /gopath/src/github.com/wercker/wercker
   steps:
+    - script:
+        name: apt-get-update
```
hatchan: We do not need this, as the install-packages already does this: https://github.com/wercker/step-install-packages/blob/master/run.sh#L6

wercker.yml
```diff
@@ -66,11 +45,12 @@ build:
     - script:
         name: gox
```
hatchan: Might as well call it `go build`

wercker.yml
```diff
@@ -66,11 +45,12 @@ build:
...
         code: |
-            gox \
+            GOOS=linux GOARCH=amd64 go build \
+            -ldflags="-X github.com/wercker/wercker/util.GitCommit=$WERCKER_GIT_COMMIT -X github.com/wercker/wercker/util.PatchVersion=$(( ($(date +%s) - $(date --date=20150101 +%s) )/(60*60*24) )) -X github.com/wercker/wercker/util.Compiled=$(date +%s)" \
+              -o "$WERCKER_OUTPUT_DIR/latest/linux_amd64/sentcli"
+            GOOS=darwin GOARCH=amd64 go build \
             -ldflags="-X github.com/wercker/wercker/util.GitCommit=$WERCKER_GIT_COMMIT -X github.com/wercker/wercker/util.PatchVersion=$(( ($(date +%s) - $(date --date=20150101 +%s) )/(60*60*24) )) -X github.com/wercker/wercker/util.Compiled=$(date +%s)" \
-              -os="linux darwin" \
-              -arch="amd64 386" \
```
hatchan: We're not building the 386 version anymore in the new version.

mwhooker:
👍

wercker.yml
termie: there is some caching going on that is breaking often

wercker.yml
termie: do we need to? i can add it back easily enough i suppose

docker/options.go
termie: yeah, kinda meh on both options, they were originally like this because they were included in the general options, but i could go remove Docker from everything everywhere i suppose

wercker.yml
termie: this doesn't appear to actually fix it (and this change is from anotehr patch), but yeah builds fail intermittently with apt-get errors

hatchan: Could you create a issues for that with the error? Maybe we can just add an `cache: false` property to the step

wercker.yml
hatchan: Don't think we publicize the URL for the 386 version. I could grab the statistics for all downloads if needed though.

Issue #78, opened 2016-02-16
Allow services to publish ports


mbrevda:
During dev/build, it can sometimes be necessary to connect to a service via it's published ports. Allow for these ports to be exposed.

It might also be useful to allow port publication to be specified by wercker.yml for the box as well as services.

/via https://werckerpublic.slack.com/archives/wercker/p1455618009000023

hatchan:
This definitely can be handy. During development it might be necessary to inspect/edit the database while developing using some UI tool. 

Currently we only link the service boxes to the main box, we should investigate if publishing these ports will give any conflicts.

mbrevda:
Any way to expedite this?

andre-hh:
+1
Is there a workaround for this?

hatchan:
Implemented in https://github.com/wercker/wercker/pull/161

PR #161, opened 2016-05-11
exposed ports for services

stvnwrgs:
See: #78

### DIFF
--- a/core/config.go
+++ b/core/config.go
@@ -38,6 +38,7 @@ type BoxConfig struct {
 	Tag        string
 	Cmd        string
 	Env        map[string]string
+	Ports      []string
 	Username   string
 	Password   string
 	Registry   string
--- a/cmd/flags.go
+++ b/cmd/flags.go
@@ -114,7 +114,9 @@ var (
 	// These flags are advanced dev settings
 	InternalDevFlags = []cli.Flag{
 		cli.BoolTFlag{Name: "direct-mount", Usage: "Mount our binds read-write to the pipeline path."},
-		cli.StringSliceFlag{Name: "publish", Value: &cli.StringSlice{}, Usage: "Publish a port from the main container, same format as docker --publish."},
+		cli.BoolFlag{Name: "expose-ports", Usage: "Enable ports from wercker.yml beeing exposed to the host system."},
+		// deprecated
+		cli.StringSliceFlag{Name: "publish", Value: &cli.StringSlice{}, Usage: "[Deprecated] Use: --expose-ports. - Publish a port from the main container, same format as docker --publish.", Hidden: true},
 		cli.BoolFlag{Name: "attach-on-error", Usage: "Attach shell to container if a step fails.", Hidden: true},
 		cli.BoolFlag{Name: "enable-volumes", Usage: "Mount local files and directories as volumes to your wercker container, specified in your wercker.yml."},
 		cli.BoolTFlag{Name: "enable-dev-steps", Hidden: true, Usage: `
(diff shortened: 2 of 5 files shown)


core/config.go
```diff
@@ -38,6 +38,7 @@ type BoxConfig struct {
 	Tag        string
 	Cmd        string
 	Env        map[string]string
+	Ports      []string
```
stvnwrgs: Needs to documented somewhere?

faiq:
Thanks for the PR @stvnwrgs! 

Looks like you reused most of the code from `docker/box.go` to do convert the port string format into a useable docker published port, which is awesome to see. However, it looks like you added a `Ports[]` field to `core/config.go` which parses the yaml. As it stands now, i believe we do port publishing through command line flag. If I'm correct this might mean that there will be two sources from which ports can be published from, which doesn't seem right to me we should probably pick one or the other.

pining @termie for more insight into the issue

stvnwrgs:
Hey @faiq, 

I thought the same about the `Ports[]`. At the moment I'm using ports in the YAML file for the services because it would be could difficult to match the ports to the right container. You would have to change the --publish functionality for that too. I also think that it might make more sense to define the ports in generally in the YAML file. When developing, it does not make sense to always add them as a param. I would suggest moving the published ports for the main container into the YAML file too. What do you think about that? With the actual PR there are no breaking changes. That's why I did it this way.

hatchan:
Auto publishing ports based on the wercker.yml should be disabled by default behind a cli flag, and maybe we should auto enable it just for `wercker dev`. 

If it will be auto enabled for `wercker dev`, then it should be possible to disable it, since the ports might already be in use.

stvnwrgs:
@hatchan I think disabling ports by default and enabling the ports automatically on the dev pipeline sounds like a good compromise. What about the command line flag? Deprecating it and configure everything over yml files?

hatchan:
Think having the commandline flag is still pretty handy. So probably document that it only applies to the main box and that it overrides the ones defined in the wercker.yml

stvnwrgs:
I think that could sound a little bit confusing for a user:

You can define ports in your yml file or overwrite them via CLI ( but only the ports for the main container). Ports for service containers are only exposed for the dev pipeline automatically.

I don't think that ports should only be enabled on the dev "pipeline", maybe you want multiple pipelines for your local development.

Another, hopefully, cleaner try:

Ports that should be exposed can be defined in the yml file. With `--expose-ports=true` (default=false) the ports are exposed. 
So the user/system can decide if it wants to export the ports in generally. 

That ports might be used already is the users problem.

faiq:
@stvnwrgs im partially in agreement with you. like i said earlier i wanted one way of doing this rather than supporting two and it seems as though supporting the syntax through the yaml makes the most amount of sense to me.

> Ports that should be exposed can be defined in the yml file. With --expose-ports=true (default=false) the ports are exposed. 
> So the user/system can decide if it wants to export the ports in generally.

Yes. I am 100% on board with this. We already established that working with multiple ports with multiple boxes/services with CLI flags would be a nightmare, and there are times when you run your pipelines where you don't need ports. In my opinion, this is a very good compromise.

@hatchan thoughts?

hatchan:
Adding the cli flag `--expose-ports` is sufficient (ie. it doesn't need true/false as an argument).

Regarding the ports flag for the main box as we currently support; we still need to support it for the time being. So adding the text that it only applies to the main box is still needed. That leaves us with the bahaviour when both are specified; if we are going to deprecate it, and both are specified then I guess we should ignore the cli flags, but it should be logged for the user.

stvnwrgs:
It is now using the ports from `--published` and showing a deprecated message.
`--expose-ports` is enabling the ports from the yml file.

Example:

```
services:
    - id: mysql
      ports:
        - "3306"


box:
  id: golang
  ports:
    - "8080"
```

hatchan:
Sorry for the wait, and thanks again for the PR. Just had to add a small fix when using the `--publish` flag: https://github.com/wercker/wercker/commit/cad414f2af67380a3e5eafe745522948b60997da

ringods:
@hatchan but when will the documentation on the devcenter site be updated?
http://devcenter.wercker.com/docs/cli/configuration/options-and-flags#-a-name-publish-class-anchor-a-publish

It's not the only case of docs that are terribly out of date.

pavel-si:
Besides, do you not think it should be done in a port-forwarding fashion, allowing the ports to be renamed? I.e., shouldn't yaml look like this:?
```
box:
  id: golang
  ports:
    - "8080":"58080"
    - "8081":"58081"
```

tamsky:
Don't y'all think it's about time to update your main docs?

http://devcenter.wercker.com/docs/cli/usage/developing still says to use

`wercker dev --publish 5000`

paulwalker:
It'd be nice if the use of `--expose-ports` would actually output something similar to what the --`--publish` flag did: `--> Forwarding localhost:3000 to 3000 on the container.`

PR #166, opened 2016-05-17
Cleanup old builds, keep latest 2

### LINKED ISSUE #165: wercker should clean up old runs locally
right now each run stays around in the .wercker dir forever, we only want maybe 2?

akupila:
This cleans up old build artifacts from `.wercker/builds`. The latest 2, sorted by modification date, are left intact.

This fixes #165

### DIFF
--- a/util/util.go
+++ b/util/util.go
@@ -25,6 +25,7 @@ import (
 	"path"
 	"path/filepath"
 	"reflect"
+	"sort"
 	"strconv"
 	"strings"
 	"sync"
@@ -408,4 +409,30 @@ func (t *Timer) Elapsed() time.Duration {
 // String repr for time
 func (t *Timer) String() string {
 	return fmt.Sprintf("%.2fs", t.Elapsed().Seconds())
+}
+
+// ByModifiedTime is used for sorting files/folders by mod time
+type ByModifiedTime []os.FileInfo
+
+// Len, Swap and Less are used for sorting
+
+// Len returns the length of items in the slice
+func (s ByModifiedTime) Len() int {
+	return len(s)
+}
+
+// Swap swaps two items when sorting
+func (s ByModifiedTime) Swap(i, j int) {
+	s[i], s[j] = s[j], s[i]
+}
+
+// Less returns true if the first value shoudl appear first in the
+// sorted results
+func (s ByModifiedTime) Less(i, j int) bool {
+	return s[i].ModTime().After(s[j].ModTime())
+}
+
+// SortByModDate sorts the files or folders descending by modification date
+func SortByModDate(dirs []os.FileInfo) {
+	sort.Sort(ByModifiedTime(dirs))
 }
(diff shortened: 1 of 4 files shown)


util/util.go
```diff
@@ -409,3 +410,21 @@ func (t *Timer) Elapsed() time.Duration {
...
 	return fmt.Sprintf("%.2fs", t.Elapsed().Seconds())
 }
+
+// ByModifiedTime is used for sorting files/folders by mod time
+type ByModifiedTime []os.FileInfo
+
+func (s ByModifiedTime) Len() int {
+	return len(s)
+}
+func (s ByModifiedTime) Swap(i, j int) {
```
hatchan: Can we have a newline between these? And they probably need a little bit of documentation as well.

util/util.go
```diff
@@ -409,3 +410,21 @@ func (t *Timer) Elapsed() time.Duration {
...
+	s[i], s[j] = s[j], s[i]
+}
+func (s ByModifiedTime) Less(i, j int) bool {
+	return s[i].ModTime().After(s[j].ModTime())
+}
+
+// SortByModDate sorts the files or folders descending by modification date
+func SortByModDate(dirs []os.FileInfo) {
```
hatchan: Maybe add a small unit test for this

hatchan:
:+1:

util/util.go
akupila: Added in `util/util_test.go`. Looks like the rest of this file is more or less boilerplate/example code?

util/util_test.go
```diff
@@ -109,3 +113,42 @@ func (s *UtilSuite) TestSplitFunc() {
 		s.Equal(len(test.output), 2)
 	}
 }
+
+func (s *UtilSuite) TestSortByModDate() {
+	// set up fixture
+	tmp, err := ioutil.TempDir("/tmp", "sort")
```
hatchan: It's not possible to create the FileInfo struct's, rather then relying on creating files on the disk

akupila: Is that a question or a statement? Not sure i understand what you mean

hatchan: question

hatchan: Isn't it possible to create a array of FileInfo objects, and then call sortByModDate on that

akupila:
Test updated to not use directories on disk (thanks Andy!). I did end up removing `WithModTime` though as it required the code outside of the tests to pass this in `[]WithModTime` too which was quite annoying as this was only added for tests. 

@hatchan @termie: "if you are running a chained build you will have a few extra builds you care about" -- what's a good solution for this? How can we know if the build is chained? Or should we only clean up dirs over a certain age (24h?)?

hatchan:
@akupila I think cleaning up only directories that are older than 24 hours is a good solution.

akupila:
@hatchan Can you have a look? It now only cleans up builds older than 24h

hatchan:
:+1:

Issue #158, opened 2016-05-04
Using a password with a $ sign breaks docker login / pulling a container


lindseybateman:

maxfrigge:
I am having the same issue with a `#`

When I save `mypass#123` my container only gets `mypass`.
Replacing the `#` sign with a `.` doesn't break and the full password gets passed.

faiq:
writing tests to reproduce the case- i am thinking its a bug in the interpolation.

faiq:
Unable to reproduce these interpolation issues that you've been having- help welcome.

hatchan:
this is caused by the way we export env vartiables. There is no escaping happening at this time, so bash interpreters it as a comment

faiq:
we interpolate things fine from the cli. including things with a `#` and a `$` I wrote a few tests in #167 to demonstrate this point. 

i think this is an escaping bug on the website

faiq:
nope @hatchan was right its bash

faiq:
i believe this particular issue is fixed with workflows because of the way we load credentials into memory. 
not fixed in the ewok stack though

kamichidu:
This bug still be alive?
In my case,  can't upload a built file to azure storage blob container `$root` .

How to reproduce:

1. Create environment variable `AzureBlobContainer` with value `$root` by wercker website.
1. Write wercker.yml referencing environment variable `AzureBlobContainer` in script step like below:
```
deploy:
  - steps:
    - script:
      name: Debug
      code: echo '$AzureBlobContainer:' "$AzureBlobContainer"
```
1. Run CI and check log.

Expected output:
```
$AzureBlobContainer: $root
```

Actual output:
```
$AzureBlobContainer: 
```

Any way escaping `$` char in environment variable?

tayzlor:
I'm also seeing this exact same thing when trying to deploy to Azure (azure has a tendency to include $ in variable values)

alphaaddynamo:
I have this problem too, how does one escape the `$` char in pipelines?

PR #171, opened 2016-05-19
attempts to solve #158

mwhooker:


mwhooker:
works how I would expect it to work:

```
../wercker/wercker/wercker --environment my.env build --pipeline build .
--> Executing pipeline
Pulling from library/alpine: latest
Already exists: d0ca440e8637
Digest: sha256:f655166f57d91bdfc8b3bc75a20391b7516de9f48ca761249c185fcb022124d2
Status: Image is up to date for alpine:latest
--> Running step: wercker-init
--> Running step: test
BAZfoo'bar
foo'bar
--> Steps passed: 2.45s
--> Pipeline finished: 3.45s
$ cat my.env
X_FOO=BAZ$BAR
X_BAR=foo'bar
$ cat wercker.yml
box:
    id: alpine:latest
    cmd: /bin/sh
build:
    steps:
        - script:
            name: test
            code: |
                echo $FOO
                echo $BAR
```

mwhooker:
Looks like it's working to me...

```
$ $GOPATH/src/github.com/wercker/wercker/wercker --environment my.env build --pipeline build .
--> Executing pipeline
Pulling from library/alpine: latest
Already exists: d0ca440e8637
Digest: sha256:f655166f57d91bdfc8b3bc75a20391b7516de9f48ca761249c185fcb022124d2
Status: Image is up to date for alpine:latest
--> Running step: wercker-init
--> Running step: test
FOO$BAR
foo'bar
--> Steps passed: 2.69s
--> Pipeline finished: 3.69s

$ cat my.env
X_FOO=FOO$BAR
X_BAR=foo'bar

$ cat wercker.yml
box:
    id: alpine:latest
    cmd: /bin/sh
build:
    steps:
        - script:
            name: test
            code: |
                echo $FOO
                echo $BAR
```

faiq:
this does make sense overall

quoting this from [here](http://mywiki.wooledge.org/Quotes)

> Single quotes: '...' removes the special meaning of every character between the quotes. The only character that you can't safely enclose in single quotes is a single quote.

and then the second part of the change is substituting any single quotes with '"'"' which ends up getting interpreted as a single quote when used later on 

👍

termie:
the substitution hurts my brain but seems on the up and up

hatchan:
Do not use single quotes, any env var in there won't be resolved in the export anymore.

termie:
i think this issue is actually up the stack from us, we want the stuff to get interpreted at this stage, it's the shell-script setting up the environment that is having issues, i'll try applying this fix to kiddie-pool instead

termie:
actually i don't know still processing

termie:
yeah i think the up the stack is correct, it looks like the thing that interprets the ENVIRONMENT file cares about "#" so quoting in the environment file is enough for part of it, but... there is some weird behavior re: variable interpretation i'm thinking a race condition on the ordering of things

termie:
like, these exports don't come out in the order expected so it is tough to do references within them, so there are two pieces to the fix:
1. quoting when we generate the environment file (i'll do that in k-pool)
2. making order deterministic if possible, 
3. towards the specific issue in #158 we have to tell them to `disable-sync: true` in the docker push step because the environment gets loaded from the shell right before that step gets parsed and the shell interprets the $ pretty much regardless of anything we do

alphaaddynamo:
this was closed but the problem was not solved and #158 is still open

PR #182, opened 2016-06-17
cleanup old builds (master)

akupila:
Exactly the same as https://github.com/wercker/wercker/pull/166 except based on master rather than ferengi


hatchan:
:+1:

PR #183, opened 2016-06-17
recommend homebrew upgrade when installed with homebrew

akupila:
- try to determine if wercker was installed with homebrew
- if yes and a new version is available recommend user to use `brew upgrade wercker-cli` instead of the built in upgrade mechanism

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -773,18 +773,18 @@ func cmdVersion(options *core.VersionOptions) error {
 				updater.ServerVersion.FullVersion())
 
 			// try to determine if binary was installed with homebrew
-			// ignore potential error; an error would mean we can't be sure
-			// so optimize for happy-path
-			homebrew, _ := util.InstalledWithHomebrew()
+			homebrew := util.InstalledWithHomebrew()
 
 			if homebrew {
 				logger.Info("\nLooks like wercker was installed with homebrew.\n\n" +
 					"To update to the latest version please use:\n" +
-					"brew upgrade wercker-cli")
-				os.Exit(1)
+					"brew update && brew upgrade wercker-cli")
+
+				logger.Println("\nUsing the built in updater can cause issues with Wercker installed with homebrew.")
+			} else {
+				logger.Infoln("Download it from:", updater.DownloadURL())
 			}
 
-			logger.Infoln("Download it from:", updater.DownloadURL())
 			if AskForUpdate() {
 				if err := updater.Update(); err != nil {
 					logger.WithField("Error", err).Warn(
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -739,6 +739,7 @@ func emitProgress(counter *util.CounterReader, total int64, logger *util.Logger)
 
 func cmdVersion(options *core.VersionOptions) error {
 	logger := util.RootLogger().WithField("Logger", "Main")
+
 	v := util.GetVersions()
 
 	if options.OutputJSON {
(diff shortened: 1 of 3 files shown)


faiq:
I think the idea is good here, but ultimately i feel 👎  about this pull request and heres why: 

1. Theres an uncomfortable amount of `exec` usage. I know that we have to use it sometimes, but we should limit executing commands on a another person's machine. 
2. A lot of string parsing. Homebrew commands and command outputs are subject to change. Relying on this is not a good idea. 

Overall, I don't feel confident that this code is very robust and it solves a little problem. Most people know how they installed wercker, so I'm assuming they can figure out that they can update it as well.

hatchan:
I don't see anything wrong with using exec, that is the only way to interact with the API of brew. 

I do agree that we shouldn't use string parsing. There is the JSON output which I think should stay stable (since it includes a version in the flag). So I think we should invoke the command, and parse the output as JSON. https://github.com/Homebrew/brew/blob/master/share/doc/homebrew/Querying-Brew.md

util/homebrew.go
```diff
@@ -0,0 +1,81 @@
...
+//
+// Heuristic (bail if any step fails):
+//   1. see if homebrew is installed
+//   2. get installed packages from homebrew, look for `wercker-cli`
+//   3. find symlink installed by homebrew, follow it to binary
+//   4. get path of currently running executable
+//   5. compare path with homebrew path
+func InstalledWithHomebrew() (bool, error) {
+	// Check if homebrew is installed
+	cmd := exec.Command("which", "brew")
```
hatchan: Don't think we need to do this. We can just try invoking brew, and if it fails just return `bool=false`

hatchan:
We can consider placing all code related to the brew JSON parsing in a opensource repo (ie. the structs)

cmd/main.go
```diff
@@ -770,6 +771,19 @@ func cmdVersion(options *core.VersionOptions) error {
 		if updater.UpdateAvailable() {
 			logger.Infoln("A new version is available:",
 				updater.ServerVersion.FullVersion())
+
+			// try to determine if binary was installed with homebrew
+			// ignore potential error; an error would mean we can't be sure
+			// so optimize for happy-path
+			homebrew, _ := util.InstalledWithHomebrew()
```
hatchan: We should probably log this error (as debug). If it isn't useful, then we probably do not want to return it from the function.

cmd/main.go
```diff
@@ -770,6 +771,19 @@ func cmdVersion(options *core.VersionOptions) error {
...
+
+			if homebrew {
+				logger.Info("\nLooks like wercker was installed with homebrew.\n\n" +
+					"To update to the latest version please use:\n" +
+					"brew upgrade wercker-cli")
+				os.Exit(1)
```
hatchan: I would prefer having a "Are you sure you want to continue" here which defaults to `no`.

util/homebrew.go
```diff
@@ -0,0 +1,81 @@
...
+	r := regexp.MustCompile(`(?m)^wercker-cli\s(.*)$`)
+	matches := r.FindAllStringSubmatch(version, -1)
+
+	if len(matches) == 0 {
+		// wercker not installed in homebrew
+		return false, nil
+	}
+
+	// resolve path to installed homebrew binary
+	cmd = exec.Command("brew", "--prefix")
```
hatchan: Are we sure we want to check if the path is the same? Maybe just checking if brew it is installed is enough

akupila:
Only parsing JSON might work but (as far as i could see) there's no info there about location of the binary. This tries to determine if the currently running binary is the one installed from homebrew. 

Of course making this simpler and just checking if homebrew is installed and if wercker was installed in homebrew might be enough. I guess it's _unlikely_ somebody would have installed wercker with homebrew but run a separately downloaded binary..

If this is the case we could simply get the JSON (should be stable according to brew docs) and look for wercker in there. If the brew is installed and wercker is found we'll assume you're using this version.

akupila:
@hatchan @faiq Now much simpler check than before. Just tries to get info about wercker from homebrew and if that fails it assumes it's not installed with homebrew.

util/homebrew.go
```diff
@@ -0,0 +1,17 @@
+package util
+
+import "os/exec"
+
+// InstalledWithHomebrew tries to determine if the cli was installed with homebrew
+func InstalledWithHomebrew() bool {
+	// Check if homebrew is installed
+	cmd := exec.Command("brew", "info", "wercker-cli")
+	err := cmd.Start()
```
hatchan: You need to use the `Run` instead of `Start`

hatchan:
:+1:

faiq:
nice and intrusive! 👍

PR #184, opened 2016-06-17
fix panic when running check-config with no box definition

davegri:
Fixes panic when trying to run the `check-config` command without a box definition

### DIFF
--- a/docker/pipeline.go
+++ b/docker/pipeline.go
@@ -44,6 +44,7 @@ func NewDockerPipeline(name string, config *core.Config, options *core.PipelineO
 	if rawBoxConfig == nil {
 		rawBoxConfig = config.Box
 	}
+	// no pipeline config or global config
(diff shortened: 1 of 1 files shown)


docker/pipeline.go
```diff
@@ -44,6 +44,9 @@ func NewDockerPipeline(name string, config *core.Config, options *core.PipelineO
 	if rawBoxConfig == nil {
 		rawBoxConfig = config.Box
 	}
+	if rawBoxConfig == nil {
```
akupila: Considering the line above is exactly the same maybe add a comment to make it clear this is not a mistake. At first glance it might look like a mistake.

faiq: the comment could be good, but looks fine to me.

faiq:
good work david! lets also get this change into master too

hatchan:
:+1:

PR #193, opened 2016-06-24
remove trailing slash from basepath (ferengi)

### LINKED ISSUE #185: Trailing slash in base-path causes the code to be put in sources subdirectory
Any trailing slashes should be removed.

akupila:
fixed #185 (ferengi)


hatchan:
:100:

PR #194, opened 2016-06-24
fix panic when trying to run an empty pipeline

davegri:


docker/pipeline.go
```diff
@@ -39,6 +39,10 @@ func NewDockerPipeline(name string, config *core.Config, options *core.PipelineO
 		return nil, fmt.Errorf("No pipeline named %s", pipelineName)
 	}
 
+	if pipelineConfig == nil {
+		return nil, fmt.Errorf("Pipeline %s is empty", pipelineName)
```
hatchan: If the pipeline contains content which doesn't match our struct this will also be nil, and thus the error message doesn't match the error anymore.

hatchan:
Small thing regarding errormessage, other than that :+1:

PR #195, opened 2016-06-24
fix panic when step is empty

davegri:


hatchan:
:+1: Don't forget to update the changelog.md

PR #197, opened 2016-06-24
Streamline deployment

hatchan:
This will add two pipelines: `release-stable` and `release-beta`. The semantics are the same as before with the deploytargets.

Both will do the following:

- Sync the binaries, version.json and checksums to our own S3 bucket (downloads.wercker.com)
- Create a new GitHub release, and add the binaries to it
- Trigger a update on the homebrew-wercker repository

There are a couple of differences between stable and beta:

- Beta adds `-beta` to the version.
- Beta GitHub release uses `prerelease`.
- Beta homebrew uses the `dev` release.

Open issues/discussion:

- Currently limited to 1 release per day. Do we want to migrate to a manual updating system?
- Ensuring that the changelog is up to date is currently not enforced.

A possible workflow is to do manual versioning:
- When a new change is created a `...-beta` version should be added to the changelog.md. Unless there is already a `-beta` version, then the change can be added to that.
  - Any pushes to `master` will be auto deployed.
  - Any existing releases will be replaced.
- When a new version is going to be released the `-beta` postfix should be removed, and it should be released. The release should enforce the version + missing `-beta` postfix.
  - This should only be possible from `master`.
  - Only allowed to release the version once.
- After a version has been released any new changes will result in a new `...-beta` version in the changelog.md and the process continues from the top.

Companion PR for homebrew-wercker: https://github.com/wercker/homebrew-wercker/pull/8


wercker.yml
```diff
@@ -66,85 +59,185 @@ build:
...
+            echo "Releasing wercker stable v$WERCKER_VERSION" > $WERCKER_REPORT_MESSAGE_FILE
+            echo "Releasing wercker stable v$WERCKER_VERSION"
+
+#    - script:
+#        name: validate changelog
+#        code: head -n 1 changelog.md | grep -q -v "v$WERCKER_VERSION"
 
     - script:
-        name: rename
+        name: rename sentcli to wercker
```
faiq: do we need this step if we name the binary properly?

hatchan: This step will go away

wercker.yml
```diff
@@ -1,57 +1,50 @@
...
 
     - script:
-        name: gox
+        name: compile
         code: |
             export CGO_ENABLED=0
             GOOS=linux GOARCH=amd64 go build \
             -ldflags="-X github.com/wercker/wercker/util.GitCommit=$WERCKER_GIT_COMMIT -X github.com/wercker/wercker/util.PatchVersion=$(( ($(date +%s) - $(date --date=20150101 +%s) )/(60*60*24) )) -X github.com/wercker/wercker/util.Compiled=$(date +%s)" \
-              -o "$WERCKER_OUTPUT_DIR/latest/linux_amd64/sentcli"
+              -o "$WERCKER_OUTPUT_DIR/artifacts/latest/linux_amd64/sentcli"
```
hatchan: Unfortunately there are still some parts that require the old naming. So think it is a better idea to do this once ewok is phased out, and as a separate PR

wercker.yml
```diff
@@ -66,85 +59,185 @@ build:
...
-            echo $WERCKER_VERSION
-            ./latest/linux_amd64/sentcli version --json > ./latest/version.json
+            export WERCKER_VERSION=$(./artifacts/latest/linux_amd64/sentcli --version | awk '{print $3}')
+            ./artifacts/latest/linux_amd64/sentcli version --json > ./artifacts/latest/version.json
+            echo "Releasing wercker stable v$WERCKER_VERSION" > $WERCKER_REPORT_MESSAGE_FILE
+            echo "Releasing wercker stable v$WERCKER_VERSION"
+
+#    - script:
+#        name: validate changelog
+#        code: head -n 1 changelog.md | grep -q -v "v$WERCKER_VERSION"
```
flenter: Can this go?

wercker.yml
```diff
@@ -66,85 +59,185 @@ build:
...
+    # Signal wercker/homebrew
+    - script:
+        name:
+        code: |
+          export DARWIN_SHA256=$(awk '{print $1}' "./artifacts/latest/darwin_amd64/SHA256SUMS")
+          curl \
+            -X POST \
+            -H "Authorization: Bearer $WERCKER_TOKEN" \
+            -H "Content-Type: application/json" \
+            --data "{ \"pipelineId\": \"5670304ed1bd8676490090fe\", \"branch\": \"update_trigger\", \"envVars\": [{ \"key\": \"CHANNEL\", \"value\": \"stable\" },{ \"key\": \"VERSION\", \"value\": \"$WERCKER_VERSION\" },{ \"key\": \"SHA256\", \"value\": \"$DARWIN_SHA256\" }] }" \
```
flenter: Maybe specify the pipelineId using an env var?

wercker.yml
```diff
@@ -66,85 +59,185 @@ build:
...
+            https://app.wercker.com/api/v3/runs
+
+release-beta:
```
flenter: Isn't this nearly completely the same as release-stable? Can't this simply be a single section in the yaml with two pipeline's (with different env-vars) in the web interface?

flenter:
Some comments, but 👍

wercker.yml
hatchan: Gone

wercker.yml
hatchan: Added

wercker.yml
hatchan: The beta is now somewhat different

PR #198, opened 2016-06-27
fix multiline env vars

akupila:
If the env vars contain a linebreak it will not come through correctly (will show up as `\n`). This should fix it by setting env vars `export KEY='multi\nline\nvalue'` (wrapped in single quotes).

1. Tested to work with something like this in an ENVIRONMENT file:
```
X_MULTILINE="this\nis\nover\nmultiple\nlines"
```

...and running the build like this:

```bash
wercker --environment ENVIRONMENT build
```

2. However this _does not_ work (result shows `\n`):
```
X_MULTILINE="this\nis\nover\nmultiple\nlines\ntoo" wercker build
```

3. This also shows `\n`:
```bash
export X_MULTILINE="this\nis\nover\nmultiple\nlines\ntoo" wercker build
wercker build
```

I'm not sure how the env vars are actually set when not running locally, might be that the solution for 1 is not sufficient?

Here's a wercker.yml to see the output:
```yml
box: golang:1.6
build:
  steps:
    - script:
        code: |
          echo "Multi:"
          echo "$MULTILINE"
```
Notice the double quote, otherwise the multiline string will be split by spaces

### DIFF
--- a/util/environment.go
+++ b/util/environment.go
@@ -74,7 +74,13 @@ func (e *Environment) Get(key string) string {
 func (e *Environment) Export() []string {
 	s := []string{}
 	for _, key := range e.Order {
-		s = append(s, fmt.Sprintf(`export %s=%q`, key, e.Map[key]))
+		val := e.Map[key]
+		if strings.Contains(val, "\n") {
+			// multiline string, need to wrap in '
+			s = append(s, fmt.Sprintf(`export %s='%s'`, key, val))
+		} else {
+			s = append(s, fmt.Sprintf(`export %s=%q`, key, val))
+		}
 	}
 	return s
 }
--- a/util/environment_test.go
+++ b/util/environment_test.go
@@ -68,3 +68,9 @@ func (s *EnvironmentSuite) TestExport() {
 	expected := []string{`export PUBLIC="foo"`, `export X_PRIVATE="zed"`}
 	s.Equal(env.Export(), expected)
 }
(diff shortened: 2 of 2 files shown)


hatchan:
Do not wrap it in single quotes, that prevents any env variables to be used inside env vars: `export MULTILINE=$OTHER`

hatchan:
`echo` has some flag that will change it interpretation of these \n characters:

```
$ export KEY='multi\nline\nvalue'
$ echo $KEY
multi
line
value

$ export KEY="multi\nline\nvalue"
$ echo $KEY
multi
line
value

$ echo -E $KEY
multi\nline\nvalue
```

termie:
i feel like this is almost certainly not wanted? the existing behavior has been in use for quite some time, if you want the value out of the env to include newlines then translate the newlines correctly with echo etc

mwhooker:
I think we need to close this without a clear use-case, and even then I think this code is going to break a lot of existing behavior

akupila:
The use-case came from https://github.com/wercker/web/pull/2098

I think it's also risky to merge this as there's a potential for breaking existing things (even if this only will change behavior when an env var contains `\n`). 

I'm wondering though what the correct way would be to pass in a multiline env var? `X_VAR="this\nis\nmultiline"`?

akupila:
Closing. Probably better to play it safe with the cli

PR #199, opened 2016-06-27
[WIP] Volumes for Services

MujibAzizi:
Allows services also to use volumes using the same format. 

This also causes the `/pipeline` directory to be created with the same contents as the main box.


MujibAzizi:
Woops, I used the wrong branch.

PR #202, opened 2016-07-01
Update test-all.sh

Tonnu:


flenter:
This fixes the test script (in that it runs the right test), but that test does fail, not?

PR #207, opened 2016-07-05
Add additional output when storing artifacts (ferengi)

### LINKED ISSUE #189: store step should explain better what it is storing
Now there's very little output:
```
Storing artifacts
Storing artifacts complete
```

We should display what we're storing in more detail: 
storing: $WERCKER_SOURCE_DIR or $WERCKER_OUTPUT_DIR
and maybe some generic info regarding nr of files/ size?

Related issue: #168

Tonnu:
Resolves #189

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,6 +1,6 @@
 ## unreleased
 
--
+- Add additional output when storing artifacts (#207)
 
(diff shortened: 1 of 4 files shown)


hatchan:
This is now ready for review again

PR #208, opened 2016-07-06
[WIP] Push to v2 registry by default

### LINKED ISSUE #180: Try to access v2 registries by default
When pushing/pulling containers, I think it makes more sense to try and connect to v2 registries first, and then, if it fails perhaps do a retry on the version of the registry.

Tonnu:
Resolves #180. 

This changes the default behaviour when pulling and pushing containers to automatically use `v2` API registries instead of `v1`. If someone needs to force `v1` they can simply append that to their registry URL, e.g:

```
- internal/docker-push:
  username: $username
  password: $password
  repository: my-registry/$repository
  registry: https://my-registry/v1
```

Todo:
- [ ] Fix tests


faiq:
Pretty good start. Should we also try to pull from a v1 registry as well? What about when the push fails, should we fall back to a v1 push?

PR #213, opened 2016-07-15
Fix failing chained runs (based on runs)

### LINKED ISSUE #151: chained workflow runs: source-dir broken
As a test I created a small test repository with a test.txt in the src folder.

And the following wercker.yml:
```
source-dir: src
box: node:0.10
build:
    steps:
        - script:
            code: |
                cat test.txt

next:
    steps:
        - script:
            code: |
                cat test.txt

final:
    steps:
        - script:
            code: |
                cat test.txt
```

The final run will fail during the wercker-init step and in the debug data of the run I can see:
`/bin/bash: line 46: cd: /pipeline/source/src: No such file or directory`

flenter:
The bug only happened when source-dir was used. The deploy like runs
would use the dir specified by source-dir as the output, causing
subsequent runs to fail because it's likely that the dir specified by
source-dir would not exist there.

The behaviour for collecting data should now be the same for deploys
as it is for builds

resolves #151

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,6 +1,7 @@
 ## unreleased
 
 - Add additional output when storing artifacts (#207)
+- fix longer (2+) chains of runs tha have source-dir specified (#151)
 
 ## v1.0.560 (2016-07-14)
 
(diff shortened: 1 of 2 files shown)


hatchan:
Please rebase, and update the changelog.md (see https://github.com/wercker/wercker/pull/207/commits/3f5bff1f51807d3ff05a5605a579e7d1eb92754a)

PR #222, opened 2016-08-05
Add support to ignore files when copying files to the container.

thomasvnoort:
This uses the library https://github.com/monochromegane/go-gitignore and supports the syntax described there.

The default location of the ignore file is .werckerignore in the project directory. It can also be specified in wercker.yml and passed to the CLI as 'ignore-file', the latter takes precedence.

If the specified file doesn't exist it is assumed to be empty and no files will be ignored.

### DIFF
--- a/core/options.go
+++ b/core/options.go
@@ -370,7 +370,7 @@ type PipelineOptions struct {
 }
 
 type PipelineDefaultsUsed struct {
-	IgnoreFileDefault bool
+	IgnoreFile bool
 }
 
 func guessApplicationID(c util.Settings, e *util.Environment, name string) string {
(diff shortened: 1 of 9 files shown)


core/options.go
```diff
@@ -364,6 +365,12 @@ type PipelineOptions struct {
 	EnableVolumes  bool
 	WerckerYml     string
 	Checkpoint     string
+
+	DefaultsUsed PipelineDefaultsUsed
+}
+
+type PipelineDefaultsUsed struct {
+	IgnoreFileDefault bool
```
termie: I would just name it IgnoreFile to match the field name, that way we can eventually automate stuff

PR #237, opened 2016-09-01
adds a run id field for easier analysis on logs

faiq:
What i think this does- adds the run id of the current build for all of the logs that will happen in the build, as we are using 1 root logger. This field will be useful for quick analysis of runs later

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -866,7 +866,7 @@ func executePipeline(cmdCtx context.Context, options *core.PipelineOptions, dock
 	// Boilerplate
 	soft := NewSoftExit(options.GlobalOptions)
 	logger := util.RootLogger().WithField("Logger", "Main")
-	logger = logger.WithField("RunID", options.RunID)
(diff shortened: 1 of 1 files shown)


cmd/main.go
```diff
@@ -866,6 +866,7 @@ func executePipeline(cmdCtx context.Context, options *core.PipelineOptions, dock
 	// Boilerplate
 	soft := NewSoftExit(options.GlobalOptions)
 	logger := util.RootLogger().WithField("Logger", "Main")
+	logger = util.RootLogger().WithField("RunID", options.RunID)
```
hatchan: Weird that this would work :-/

You could also use `WithFields` (note the `s`). Might make more sense when you see this for the first time.

faiq: I think the logger is a singleton, although I wish i could set the `WithField` option globally so i get every single log.... right now it seems like only top level logs are showing up. like pipeline started or step started or step finished... but maybe thats all we need after all?

PR #238, opened 2016-09-02
Fix injection of bash into empty box's cmd with overridden entrypoint

### LINKED ISSUE #218: wercker should not inject a cmd when entrypoint is used
Currently we inject `/bin/bash` when no `cmd` is provided for a box. This behavior is not desired when overriding the entrypoint (if the container provides an alternative entrypoint):

```
push-quay:
  box:
    id: quay.io/wercker/cronetes
    entrypoint: /bin/sh
```

See https://github.com/wercker/wercker/blob/master/docker/box.go#L87-L92

kokaz:
This add extra check on `cmd` and `entrypoint` in order to have proper behaviour when overriding the entrypoint with an empty `cmd`. This should work:
```
push-quay:
  box:
    id: quay.io/wercker/cronetes
    entrypoint: /bin/sh
```

Fix #218

### DIFF
--- a/docker/box.go
+++ b/docker/box.go
@@ -84,12 +84,12 @@ func NewDockerBox(boxConfig *core.BoxConfig, options *core.PipelineOptions, dock
 
 	networkDisabled := false
 
+	entrypoint := boxConfig.Entrypoint
+
 	cmd := boxConfig.Cmd
-	if cmd == "" {
+	if entrypoint == "" && cmd == "" {
 		cmd = "/bin/bash"
 	}
-
-	entrypoint := boxConfig.Entrypoint
 
 	logger := util.RootLogger().WithFields(util.LogFields{
 		"Logger":    "Box",


docker/box.go
```diff
@@ -84,17 +84,19 @@ func NewDockerBox(boxConfig *core.BoxConfig, options *core.PipelineOptions, dock
...
-	entrypoint := boxConfig.Entrypoint
-
 	logger := util.RootLogger().WithFields(util.LogFields{
-		"Logger":    "Box",
-		"Name":      name,
-		"ShortName": shortName,
+		"Logger":     "Box",
+		"Name":       name,
+		"ShortName":  shortName,
+		"Entrypoint": entrypoint,
```
hatchan: Don't think we have to add these to the logger, as this will be added to each log statement.

hatchan:
Minor comment, other than that :lgtm:

docker/box.go
kokaz: My bad added it for debugging purposes

kokaz:
@hatchan shouldn't we change the default `cmd` from `bash` to `sh` because it could be an issue for containers based on `alpine` for example.

hatchan:
@kokaz This will have more impact, which needs more extensive testing. Best to create a separate PR for that, and leave this one as is.

kokaz:
@hatchan is it mergeable?

PR #240, opened 2016-09-16
Add support to ignore files when copying files to the container.

termie:
* Add support to ignore files when copying files to the container.
This uses the library https://github.com/monochromegane/go-gitignore and supports the syntax described there.
The default location of the ignore file is .werckerignore in the project directory. It can also be specified in wercker.yml and passed to the CLI as 'ignore-file', the latter takes precedence.
If the specified file doesn't exist it is assumed to be empty and no files will be ignored.
* Rename IgnoreFileDefault to IgnoreFile.


termie:
ferengi version of this

PR #242, opened 2016-09-22
update to go 1.7

termie:
fixes support for sierra


wercker.yml
```diff
@@ -91,6 +91,27 @@ run-tests:
               cd $CODE_PATH
               ./test-all.sh
 
+wercker-in-a-box:
```
hatchan: This has nothing to do with the upgrade to 1.7

changelog.md
```diff
@@ -1,5 +1,11 @@
 ## unreleased
 
+## v1.0.630 (2016-09-22)
```
hatchan: This is outdated

JeanMertz:
@hatchan @termie I'm in the dark on this one.

I see a lot of (crucial) fixes merged to the `ferengi` branch (I guess "the next big release"?), what about us using Sierra and needing a working client binary right now?

I've got quite some developers at our office complaining about Wercker not working on Sierra, and right now we're sharing a self-compiled binary that we're hosting ourselves.

Is that the way to go for now (which works, but isn't exactly user-friendly), or is there some other release process you use?

hatchan:
@JeanMertz We're merging the ferengi branch really soon. We had to do this because we maintained two stacks at the same time. 

I do not have a Mac so cannot help with you on that one. Could you create a small issue explaining the issue, and your workaround?

PR #250, opened 2016-10-11
chmod the /pipeline/source dir

termie:
something in the new kubernetes version of the builds is making this directory only readable only by root, some stuff like nginx and apache operate as nobody


termie:
+1'ing myself

PR #251, opened 2016-10-11
force http1 on go1.7

termie:
updates two deps to use a force_http1 branch


termie:
+1'ing myself also, gonna test this more on staging, worked locally (and failed before i made this change)

PR #271, opened 2017-01-06
fix volumes bindings for services

ankit-arora:
I added a fix for persistent volume binding for services. Helped me in mounting a host directory to /data/db directory for my mongo service for dev pipeline.


termie:
hi there, this looks like it'll do the job, @faiq is writing a test for it right now

faiq:
test done accepting :) thanks for putting this together :)

PR #272, opened 2017-01-09
Remove auth token temporarily when getting a step

kokaz:
The client used to fetch steps is only used for that purpose, as this call doesn't need to be authenticated I removed the token which were associated to this client so that if the token is invalid or missing the step fetching will still work.


hatchan:
This has some concurrency issues. Isn't it possible to either use a different client for this, or allow the caller to temporary disable the auth?

kokaz:
@hatchan You're right, I'll investigate on alternative solutions

kokaz:
@hatchan I changed the approach, see edited PR :)

core/step.go
```diff
@@ -355,9 +355,9 @@ func (s *ExternalStep) Fetch() (string, error) {
 
 			// TODO(termie): probably don't need these in global options?
 			apiOptions := api.APIOptions{
-				BaseURL:   s.options.GlobalOptions.BaseURL,
-				AuthToken: s.options.GlobalOptions.AuthToken,
+				BaseURL: s.options.GlobalOptions.BaseURL,
 			}
+			// WARNING: this client doesn't contain any auth token
```
hatchan: One nitpicky thing: please use `NOTE(kokaz): ...`

kokaz: @hatchan fixed :)

PR #275, opened 2017-01-20
Azure support

faiq:
- adds a new `Authenticatable` for Azure
- checks for errors if `Authenticator` wasn't able to be made
- adds test

### DIFF
--- a/auth/auth_helper.go
+++ b/auth/auth_helper.go
@@ -10,15 +10,22 @@ import (
 )
 
 type CheckAccessOptions struct {
-	Username      string
-	Password      string
-	AwsSecretKey  string
-	AwsAccessKey  string
-	AwsRegion     string
-	AwsStrictAuth bool
-	AwsID         string
-	AwsRegistryID string
-	Registry      string
+	Username               string
+	Password               string
+	AwsSecretKey           string
+	AwsAccessKey           string
+	AwsRegion              string
+	AwsStrictAuth          bool
+	AwsID                  string
+	AwsRegistryID          string
+	Registry               string
+	AzureClientID          string
+	AzureClientSecret      string
+	AzureSubscriptionID    string
+	AzureTenantID          string
+	AzureResourceGroupName string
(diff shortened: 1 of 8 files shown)


core/config_test.go
```diff
@@ -74,23 +74,23 @@ func (s *ConfigSuite) TestConfigBoxStructs() {
 	_, ok := amzn.Box.Auth.(*AmazonAuth)
 	s.Equal(ok, true)
 	env := util.NewEnvironment(os.Environ()...)
-	authenticator := amzn.Box.Auth.ToAuthenticator(env)
+	authenticator, _ := amzn.Box.Auth.ToAuthenticator(env)
```
hatchan: Probably a good idea to check the error if it is `nil`

auth/auth_helper.go
```diff
@@ -73,6 +80,10 @@ func GetRegistryAuthenticator(opts CheckAccessOptions) (auth.Authenticator, erro
 		return auth.NewAmazonAuth(opts.AwsRegistryID, opts.AwsAccessKey, opts.AwsSecretKey, opts.AwsRegion, opts.AwsStrictAuth), nil
 	}
 
+	if opts.AzureClientID != "" && opts.AzureClientSecret != "" && opts.AzureSubscriptionID != "" && opts.AzureTenantID != "" && opts.AzureResourceGroupName != "" && opts.AzureRegistryName != "" && opts.AzureLoginServer != "" {
```
hatchan: This is a pretty long line, maybe make it a multiline or define the logic in a separate func (possibly defined on `opts`)

core/config_test.go
faiq: sure! probably should add the azure stuff in the tests too

auth/auth_helper.go
faiq: i don't really think this is that big of a deal, but if you want i can refactor this and the amazon stuff in a different pull request

termie:
:+1:

PR #276, opened 2017-01-24
helper options maker

faiq:
- moves the options maker to a helper function

### DIFF
--- a/docker/docker.go
+++ b/docker/docker.go
@@ -774,73 +774,88 @@ func (s *DockerPushStep) InitEnv(env *util.Environment) {
 	}
 
 	//build auther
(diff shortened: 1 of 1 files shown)


termie:
hmm, not quite what i was looking for, i meant to write a function like
```
function optHelper(data, key string) string {
  if s, ok := data[key]; ok {
    return env.Interpolate(username)
  }
  return ""
}
```

and replace all those calls with oneliners:

`opts.Username = optHelper(data, "username")`

termie:
(and to do that for all the places that do that in our codebase)

termie:
is also still fine to put these in a separate opts struct and whatnot, the main readability improvement is getting rid of those repeated lines of code with two words changed

PR #278, opened 2017-01-26
update change log

faiq:


faiq:
verbal +1 from @termie

PR #296, opened 2017-04-06
allow relative paths for file:// dev steps

termie:
and symlink them for rapid testing


termie:
merging this as it is an undocumented feature anyway

PR #298, opened 2017-04-07
remove docker prefix from docker opts

termie:
some pre-cleanining before i add some more docker bits


termie:
strictly a go-rename thing so all works, merging

PR #301, opened 2017-04-25
automatically detect bash or sh

termie:
# wat 

- Automatically detect bash or sh for containers by default,
defaulting to bash if it is there (#301)
- Fix a small bug when doing local deploys and using a working-dir other
 than .wercker (#301)

# y 

alpine containers use /bin/sh, many others use /bin/bash, our default has always been /bin/bash, but this made many containers fail until a command was set explicitly, this removes that need

re deploy bug, local deploy auto-path-lookup-magic was using a hardcoded .wrcker instead of looking at working-dir

### DIFF
--- a/docker/box.go
+++ b/docker/box.go
@@ -87,7 +87,7 @@ func NewDockerBox(boxConfig *core.BoxConfig, options *core.PipelineOptions, dock
 
 	cmd := boxConfig.Cmd
 	if cmd == "" {
-		cmd = "/bin/bash"
+		cmd = DefaultDockerCommand
 	}
 
 	entrypoint := boxConfig.Entrypoint
--- a/docker/box.go
+++ b/docker/box.go
@@ -331,7 +331,10 @@ func (b *DockerBox) RecoverInteractive(cwd string, pipeline core.Pipeline, step 
 	env = append(env, pipeline.Env().Hidden.Export()...)
 	env = append(env, step.Env().Export()...)
 	env = append(env, fmt.Sprintf("cd %s", cwd))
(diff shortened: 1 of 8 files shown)


docker/box.go
```diff
@@ -331,7 +331,7 @@ func (b *DockerBox) RecoverInteractive(cwd string, pipeline core.Pipeline, step
 	env = append(env, pipeline.Env().Hidden.Export()...)
 	env = append(env, step.Env().Export()...)
 	env = append(env, fmt.Sprintf("cd %s", cwd))
-	cmd := []string{b.cmd}
+	cmd, _ := shlex.Split(b.cmd)
```
jwolski2: Why ignore the error?

tests/projects/scratch-n-push/wercker.yml
```diff
@@ -10,5 +10,5 @@ build:
         username: $DOCKER_USER
         password: $DOCKER_PASSWORD
         registry: https://quay.io
-        tag: $WERCKER_GIT_BRANCH-$WERCKER_GIT_COMMIT
+        tag: uniqueTagFromTest
```
jwolski2: Was this an accidental modification?

termie: i think there was an accidental modification that made this change a long time ago, the correct tag is the uniqueTagFromTest which is checked for in the test

docker/box.go
termie: ah, yeah here we shouldn't ignore, we ignore it in the other places because we are actually parsing a static string in those places

PR #302, opened 2017-04-25
Creates a canary deploy endpoint

faiq:
What: 
- adds a post request to a canary service with the current git commit and branch

Why: 
- we should be able to deploy a release of this canary from here


faiq:
for the time being, I'm going to base this #301 because i want to use that experiment

wercker.yml
```diff
@@ -205,3 +205,29 @@ release-beta:
...
+        source-dir: "./artifacts"
+        delete-removed: false
+        bucket-url: $AWS_BUCKET_URL/canary-$WERCKER_GIT_BRANCH/
+        key-id: $AWS_ACCESS_KEY_ID
+        key-secret: $AWS_SECRET_ACCESS_KEY
+
+    - script:
+        name: Create canary deploy
+        code: |
+          data="{\"sentcliBranch\": \"canary-$WERCKER_GIT_BRANCH\", \"sentcliCommit\": \"$WERCKER_GIT_COMMIT\", \"users\": [ \"$CANARY_USERS\" ], \"percentage\": $CANARY_PERCENTAGE, \"total\": $CANARY_TOTAL}"
```
termie: @faiq note the env var changes, this isn't a staging-specific step, we'll target prod or staging by changing the CANARY_ENDPOINT below

wercker.yml
```diff
@@ -205,3 +205,29 @@ release-beta:
...
+
+canary:
+  steps:
+    - script:
+        name: prepare
+        code: |
+            echo $WERCKER_GIT_COMMIT > ./artifacts/HEAD
+            cp -r ./artifacts/latest ./artifacts/$WERCKER_GIT_COMMIT
+
+    - s3sync:
```
termie: @faiq we deploy our builds to a special branch canary-$BRANCH and use that for the tests, this way you don't have to chain the canary pipeline from a release pipeline and use it directly from a build instead

faiq: MUCH BETTER. YES THANK U.

real talk i was wondering why all of my shit was failing and it was because of no s3 sync thanks for updating

wercker.yml
faiq: yeah i was on the struggle bus with curl and an envvar wasn't working, probably a quoting issue.

PR #303, opened 2017-04-26
Update reporter-client

hatchan:
# What
Update `reporter-client` dependency to `master`

# Why
The feature branch was merged into master


termie:
yayy

PR #305, opened 2017-05-10
fix parameter not to be string in canary

faiq:
#304 made it fast fail a string, when its actually a number. this fixes that


termie:
çmsb

faiq:
works fine https://app.wercker.com/wercker/wercker/runs/canary-staging-wercker/591386cb2add12000158730f?step=591386d169ca6c00014444f7 merging

Issue #306, opened 2017-05-23
Panic when Docker repository invalid


jwolski2:
The following happens if you reference a box in your pipeline that points to an invalid Docker repository or register:

```
--> No Docker host specified, checking: /var/run/docker.sock
--> Executing pipeline
--> Running step: setup environment
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0x745b32]

goroutine 1 [running]:
panic(0x9d66e0, 0xc42000e0a0)
    /usr/local/go/src/runtime/panic.go:500 +0x1a1
github.com/wercker/wercker/vendor/github.com/wercker/docker-check-access.DockerAuth.Repository(0x0, 0x0, 0xc42013f180, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x545a20, ...)
    /go/src/github.com/wercker/wercker/vendor/github.com/wercker/docker-check-access/docker-auth.go:126 +0x42
github.com/wercker/wercker/vendor/github.com/wercker/docker-check-access.DockerAuthV1.Repository(0xc420323380, 0x0, 0x0, 0x0, 0x0)
    <autogenerated>:25 +0x78
github.com/wercker/wercker/docker.(*DockerBox).Fetch(0xc4200e84b0, 0x7f9c039f8708, 0xc4202c5a70, 0xc42031cae0, 0x0, 0x0, 0x0)
    /go/src/github.com/wercker/wercker/docker/box.go:527 +0x14f
github.com/wercker/wercker/cmd.(*Runner).SetupEnvironment(0xc420326370, 0x7f9c039f8708, 0xc4202c5a70, 0x0, 0x0, 0x0)
    /go/src/github.com/wercker/wercker/cmd/runner.go:530 +0x7af
github.com/wercker/wercker/cmd.executePipeline(0x7f9c039f8708, 0xc4202c5a70, 0xc4202d1800, 0xc420326000, 0xc420324840, 0x0, 0x0, 0x0)
    /go/src/github.com/wercker/wercker/cmd/main.go:937 +0x7e2
github.com/wercker/wercker/cmd.cmdBuild(0x7f9c039f7538, 0xc42006c398, 0xc4202d1800, 0xc420326000, 0xc42006c398, 0x0, 0xc420302b76)
    /go/src/github.com/wercker/wercker/cmd/main.go:413 +0xb0
[...]

PR #308, opened 2017-05-23
Throw error for pulling a blank repo

faiq:
quick check for a repository before trying to get the authenticator's version of it 

fixes the panic described in #306

### DIFF
--- a/docker/box.go
+++ b/docker/box.go
@@ -547,6 +547,10 @@ func (b *DockerBox) Fetch(ctx context.Context, env *util.Environment) (*docker.I
 	if err != nil {
 		return nil, err
 	}
+	rep := env.Interpolate(b.repository)
+	if rep == "" {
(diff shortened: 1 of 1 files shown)


faiq:
The only other place we do this is in the push step, and after trying an example we don't get a panic we get a error that says invalid reference format

```
  7 build:
  8   steps:
  9     - script:
 10       code: |
 11         echo "hello world"
 12     - internal/docker-push:
 13         repository:
 14         username: _json_key
 15         password: $GCR_JSON
 16         registry: gcr.io/v2
 17         cmd: /bin/sh
```

error: 
```
--> Running step: docker push
WARNING Registry address is expected to begin with 'https://', forcing it to use https
ERROR Error interacting with this repository:  invalid reference format
--> Step failed: docker push 0.01s
--> Pipeline failed: 10.27s
FATAL Step failed: docker push
```

So it looks like we're good

PR #311, opened 2017-06-09
Use new steps registry

mrolla:
Adds support for querying and fetching steps from the new registry.


core/step.go
```diff
@@ -461,9 +466,15 @@ func (s *ExternalStep) Execute(sessionCtx context.Context, sess *Session) (int,
 		}
 	}
 
-	if yes, _ := util.Exists(s.HostPath("run.sh")); yes {
-		exit, _, err := sess.SendChecked(sessionCtx, fmt.Sprintf(`source "%s" < /dev/null`, s.GuestPath("run.sh")))
+	// If step.sh exists run it if not try falling back to run.sh
+	if yes, _ := util.Exists(s.HostPath("step.sh")); yes {
```
jwolski2: Why ignore the `err` coming back from the call to `Exists()`?

PR #313, opened 2017-06-19
Fix invalid symlink after extracting artifacts

### LINKED ISSUE #300: Symlinks aren't passed through Pipelines
With the following wercker.yml:

```
box:
  id: alpine
  cmd: /bin/sh
build:
  steps:
    - script:
        name: Write some files
        code: echo "test content" | tee one two three
    - script:
        name: Create a symlink
        code: ln -s three four
    - script:
        name: See what we've got...
        code: ls -lth

deploy:
  steps:
    - script:
        name: What do we have now?
        code: |
          ls -lth
```

Expected behaviour: Symlinks should be intact
Actual behaviour: A zero-size file is left in the symlink's place.

```
-> % wercker build --artifacts && wercker deploy
--> Executing pipeline
Pulling from library/alpine: latest
Digest: sha256:58e1a1bb75db1b5a24a462dd5e2915277ea06438c3f105138f97eb53149673c4
Status: Image is up to date for alpine:latest
--> Running step: wercker-init
--> Running step: Write some files
test content
--> Running step: Create a symlink
--> Running step: See what we've got...
total 28
lrwxrwxrwx    1 root     root           5 Apr 21 17:28 four -> three
-rw-r--r--    1 root     root          13 Apr 21 17:28 one
-rw-r--r--    1 root     root          13 Apr 21 17:28 three
-rw-r--r--    1 root     root          13 Apr 21 17:28 two
-rw-r--r--    1 root     root         396 Apr 21 17:28 wercker.yml
Storing artifacts
Collecting files from /pipeline/source
./.git/ (content omitted)
./four
./one
./three
./two
./wercker.yml

Total artifact size: 32 MiB
Storing artifacts complete
--> Steps passed: 118.94s
--> Pipeline finished: 119.94s
No target specified, using recent build output.
--> Executing pipeline
Pulling from library/alpine: latest
Digest: sha256:58e1a1bb75db1b5a24a462dd5e2915277ea06438c3f105138f97eb53149673c4
Status: Image is up to date for alpine:latest
--> Running step: wercker-init
--> Running step: What do we have now?
total 28
-rw-r--r--    1 root     root          13 Apr 21 17:30 one
-rw-r--r--    1 root     root          13 Apr 21 17:30 three
-rw-r--r--    1 root     root          13 Apr 21 17:30 two
-rw-r--r--    1 root     root         396 Apr 21 17:30 wercker.yml
-rwxr-xr-x    1 root     root           0 Apr 21 17:29 four
--> Steps passed: 101.60s
--> Pipeline finished: 102.61s
```

Confirmed the issue is reproducible through the CLI and on Wercker web. A common workaround is to tar up the contents of $WERCKER_OUTPUT_DIRECTORY at the end of the first pipeline, and untar at the beginning of the second, but this is less than ideal.

 https://app.wercker.com/riceo/wercker-sandbox/runs/build/58fa3f5f8820930001535047 & https://app.wercker.com/riceo/wercker-sandbox/runs/deploy/58fa3f6cdd167c010099fcb1

kokaz:
When downloading then extracting artifacts symlinks where not correctly extracted. This is because in `archive.go` there were no special handling of symlinks.

Fix #300


util/archive.go
```diff
@@ -238,21 +238,30 @@ func (p *ArchiveExtract) Process(hdr *tar.Header, r io.Reader) (*tar.Header, io.
...
 
-	// Extract the file!
-	file, err := os.OpenFile(fpath, os.O_WRONLY|os.O_CREATE, hdr.FileInfo().Mode())
-	if err != nil {
-		return hdr, r, err
-	}
-	defer file.Close()
+	switch hdr.Typeflag {
+	case tar.TypeDir:
+		if _, err := os.Stat(fpath); err != nil {
```
hatchan: Why add this os.Stat check here? Can't you just do the MkdirAll and if there is an error check if it is `os.IsExist`

kokaz: `MkdirAll` will do a `Stat` but will fail only if `Stat` returns that `fpath` is regular file. 
So in this case if a file is already there with the same name as the directory we want to create it will not fail. I think I'll just remove this "guard" as it does more harm than good.

PR #314, opened 2017-06-22
Keep the order when collecting pass-through vars

reclaro:
The GetPassthru and GetHiddenPassthru functions don't keep the order of
the Envs, that is because they range over the Environment.Map. The fix
is to range over the Environment.Order.
In addition to that we added a small refactor to put the logic of the
collection mechanism in a single function so we can avoid code
duplication.

### DIFF
--- a/util/environment.go
+++ b/util/environment.go
@@ -20,6 +20,11 @@ import (
 	"strings"
 )
 
+var (
+	protected = "XXX_"
+	public    = "X_"
+)
+
 // Environment represents a shell environment and is implemented as something
 // like an OrderedMap
 type Environment struct {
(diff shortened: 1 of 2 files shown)


reclaro:
To reproduce the issue:
Just add one of the 2 tests added in this PR and then run tests several times and you'll see that 
eventually they will fail:
```bash
for i in  {1..50}; do go test; done                                                                                                                                                                                                                                                                                                                      
PASS
ok      github.com/wercker/wercker/util 0.015s
PASS
ok      github.com/wercker/wercker/util 0.014s
PASS
ok      github.com/wercker/wercker/util 0.014s
PASS
ok      github.com/wercker/wercker/util 0.015s
PASS
ok      github.com/wercker/wercker/util 0.014s
--- FAIL: TestPasstrhuKeepsOrder (0.00s)
        Error Trace:    environment_test.go:76
        Error:          Not equal:
                        expected: []string{"fake1", "fake2", "fake3"}
                        received: []string{"fake3", "fake1", "fake2"}

                        Diff:
                        --- Expected
                        +++ Actual
                        @@ -1,5 +1,5 @@
                         ([]string) (len=3) {
                        + (string) (len=5) "fake3",
                          (string) (len=5) "fake1",
                        - (string) (len=5) "fake2",
                        - (string) (len=5) "fake3"
                        + (string) (len=5) "fake2"
                         }
--- FAIL: TestEnvironmentSuite (0.00s)
FAIL
exit status 1
```

util/environment.go
```diff
@@ -103,28 +108,29 @@ var mirroredEnv = [...]string{
...
+	return e.passthru(public)
 }
 
 // Collect the hidden passthru variables
 func (e *Environment) GetHiddenPassthru() (env *Environment) {
+	return e.passthru(protected)
+}
+
+func (e *Environment) passthru(envType string) (env *Environment) {
+	prefix := "X_"
```
hatchan: Maybe we can remove this prefix from `passthru` and have it be a parameter instead of `envType`. The two constants can then contain `X_` and `XXX_`

reclaro: Indeed, new PR coming

util/environment_test.go
```diff
@@ -39,6 +39,19 @@ func (s *EnvironmentSuite) TestPassthru() {
 	s.Equal(1, len(env.GetHiddenPassthru().Ordered()))
 }
 
+func (s *EnvironmentSuite) TestPasstrhuKeepsOrder() {
```
mrolla: Small typo

util/environment_test.go
```diff
@@ -39,6 +39,19 @@ func (s *EnvironmentSuite) TestPassthru() {
...
+	env := NewEnvironment("X_fake1=val1", "X_fake2=val2", "X_fake3=$fake2")
+	actual := env.GetPassthru()
+	expected := []string{"fake1", "fake2", "fake3"}
+	s.Equal(expected, actual.Order)
+}
+func (s *EnvironmentSuite) TestPasstrhuHiddenKeepsOrder() {
```
mrolla: See above

util/environment_test.go
reclaro: done

PR #315, opened 2017-06-22
fix: (wercker/envvars#55)  Remove godotenv dep and load env file in o…

simonlord:
…rder ourselves

godotcom was loading the environment file into a map and iterating  over that map to add the env key/val pairs back into the processes environment (through os.Setenv).

Iterating over a map is a randomized order in golang.

So if you have "NAME=Simon" and "GREETING=Hello $NAME" in the file then sometimes you'd get "GREETING=Hello " and sometimes it'd be "GREETING=Hello Simon" depending on the randomized order when iterating over the map.

So i've added a simplified file loading function to util.Environment

Add test too.


simonlord:
Just an FYI - but this is needed by https://github.com/wercker/wercker/pull/314 to complete https://github.com/wercker/envvars/issues/55

util/environment.go
```diff
@@ -155,3 +156,34 @@ func (e *Environment) GetInclHidden(key string) string {
 
 	return ""
 }
+
+// LoadFile imports key,val pairs from the provided file path. File entries
+// should be 1 per line in the form key=value. Blank lines and lines begining
+// with # are ignored.
+func (e *Environment) LoadFile(f string) error {
```
reclaro: nit: do you think LoadFromFile explains better what is the func job?

util/environment_test.go
```diff
@@ -68,3 +68,21 @@ func (s *EnvironmentSuite) TestExport() {
...
+		[]string{"C", "3"},
+		[]string{"D", "4"},
+		[]string{"E", "5"},
+		[]string{"F", "6"},
+		[]string{"G", "7"},
+	}
+	env.LoadFile("../tests/environment-test-load-file.env")
+	s.Equal(8, len(env.Ordered()), "Should only load 8 valid lines.")
+	s.Equal("foo", env.Get("PUBLIC"), "LoadFile should ignore keys already set in env.")
+	s.Equal(expected, env.Ordered(), "LoadFile should maintain order.")
```
reclaro: Not sure about that last assert, shall we actually check that the env.Order contains the list of the key in the expected order?

simonlord: `s.Equal()` is performing `reflect.DeepEqual(expected, actual)` so checks each item of the slice as it iterates through it.

reclaro: Thanks for your answer, I think I haven't explain myself very well. MY point is that we should assert directly the expected with the Order field of the Environment struct as that field is the field which keeps the ordering and it is used is several other functions. The Ordered functions use that field but it could go away (according to the comment it is just for debugging even if I noticed that it is used). In other words I'd prefer to assert the correct ordering of the struct without passing via an intermediate function.

util/environment.go
simonlord: I think they're about the same, and shorter names are better.

util/environment_test.go
simonlord: Yes good point, i've added an assert for that now too 👍

util/environment.go
```diff
@@ -155,3 +156,34 @@ func (e *Environment) GetInclHidden(key string) string {
...
+	file, err := os.Open(f)
+	if err != nil {
+		return err
+	}
+	defer file.Close()
+
+	s := bufio.NewScanner(file)
+	for ok := s.Scan(); ok; ok = s.Scan() {
+		// Ignore comments
+		if strings.Index(s.Text(), "#") == 0 {
```
mrolla: Consider using `strings.HasPrefix`. It is slightly more idiomatic and self explanatory.

simonlord: Good call @mrolla - i'll do that now

PR #317, opened 2017-06-26
fix: Handle quoted environment variable values in environment file

simonlord:
Missed the fact that godotenv handled quotes and that wercker environment files are quoted. This fixes that.

### DIFF
--- a/util/environment.go
+++ b/util/environment.go
@@ -184,8 +184,30 @@ func (e *Environment) LoadFile(f string) error {
 		if e.Get(key) != "" {
 			continue
 		}
+
+		val = trim(val)
+
 		e.Add(key, val)
 	}
 
 	return nil
+}
+
+func trim(s string) string {
+	s = strings.TrimSpace(s)
+
+	if len(s) > 1 {
+		f := string(s[0:1])
+		l := string(s[len(s)-1:])
+		if f == l && strings.ContainsAny(f, `"'`) {
+			// strip surrounding quotes
+			s = string(s[1 : len(s)-1])
+
+			// now expand escaped double quotes and newlines
+			s = strings.Replace(s, `\"`, `"`, -1)
+			s = strings.Replace(s, `\n`, "\n", -1)
+		}
+	}
+
+	return s
 }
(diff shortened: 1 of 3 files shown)


util/environment.go
```diff
@@ -184,8 +184,30 @@ func (e *Environment) LoadFile(f string) error {
...
 		e.Add(key, val)
 	}
 
 	return nil
 }
+
+func trim(s string) string {
+	s = strings.Trim(s, " ")
+
+	if len(s) > 1 {
```
mrolla: A string with just `'` or `"` would end up populating a variable with just a single/double quote. This is definitely an user error, but it seems weird to me to have that anyway. Thoughts?

util/environment.go
```diff
@@ -184,8 +184,30 @@ func (e *Environment) LoadFile(f string) error {
...
+		val = trim(val)
+
 		e.Add(key, val)
 	}
 
 	return nil
 }
+
+func trim(s string) string {
+	s = strings.Trim(s, " ")
```
hatchan: Not sure if we should trim, what if the value that the users provides is an empty string.

util/environment.go
simonlord: Yeah that's a tricky one. Definitely user error. They should escape it themselves. I guess the export would fail and the return code would bubble up and show an error in the setup environment step?

util/environment.go
mrolla: `strings.TrimSpace`

util/environment_test.go
```diff
@@ -93,10 +93,15 @@ func (s *EnvironmentSuite) TestLoadFile() {
 		[]string{"E", "5"},
 		[]string{"F", "6"},
 		[]string{"G", "7"},
+		[]string{"H", "8"},
+		[]string{"I", "9"},
+		[]string{"J", "Hello \"quotes\""},
+		[]string{"K", ""},
+		[]string{"L", "\n"},
```
mrolla: Can you try adding these two?
```
[]string{"M", `\"`},
[]string{"N", `"`},
```

simonlord: Added, and they pass the test.

PR #325, opened 2017-08-07
Allow docker-push step to seemlessly push to a default registry

tjames221188:
We now try to infer the registry from the repository, previously the user would have to specify:

```yaml
registry: https://quay.io
repository: quay.io/user/app
```

If a user don't specify a registry and the repository contains another registry than the Docker Hub one we automatically use the v2 of the registry.


docker/docker.go
```diff
@@ -838,15 +843,34 @@ func (s *DockerPushStep) InitEnv(env *util.Environment) {
...
 		opts.Password = s.options.AuthToken
-		s.repository = fmt.Sprintf("%s/%s/%s", s.options.WerckerContainerRegistry.Host, s.options.ApplicationOwnerName, s.options.ApplicationName)
 		s.builtInPush = true
 	}
-	auther, _ := dockerauth.GetRegistryAuthenticator(opts)
 
+	return opts
+}
+
+// The IStep Interfacew
```
faiq: typo

hatchan: Looks like this can be removed

docker/docker.go
```diff
@@ -43,6 +43,8 @@ import (
 	"github.com/wercker/wercker/core"
 	"github.com/wercker/wercker/util"
 	"golang.org/x/net/context"
+	"github.com/docker/distribution/reference"
+	"net/url"
```
hatchan: Can you move the stdlib imports to the list above. I recommend using goimports which will do these things for you.

docker/docker_push_test.go
```diff
@@ -7,6 +7,7 @@ import (
 	"github.com/stretchr/testify/suite"
 	"github.com/wercker/wercker/core"
 	"github.com/wercker/wercker/util"
+	"fmt"
```
hatchan: Move to the list above

docker/docker.go
tjames221188: Typo fixed. @hatchan I actually found the comment useful - it's nice to know that everything below this line implements the `Step` interface. Unless you strongly disagree, I'd like to leave it in.

docker/docker.go
tjames221188: done

docker/docker_push_test.go
tjames221188: done

docker/docker.go
hatchan: Thought it was a comment for something which was removed. If it applies to everything below, then I don't mind.

docker/docker.go
```diff
@@ -838,15 +842,34 @@ func (s *DockerPushStep) InitEnv(env *util.Environment) {
...
+		domain := reference.Domain(x)
+
+		if domain != "docker.io" {
+			reg := &url.URL{Scheme: "https", Host: domain, Path: "/v2"}
+			opts.Registry = reg.String() + "/"
+		}
+	}
+
+	// Set user and password automatically if using wercker registry
+	if opts.Registry == s.options.WerckerContainerRegistry.String()+"/v2/" {
```
jwolski2: @kokaz and I noticed a bug here. "/v2" must not be appended to the WerckerContainerRegistry value. @kokaz fixed this bug in another commit and will cherry-pick that commit into this branch.

kokaz:
PR updated see 571ef3d

PR #326, opened 2017-08-08
Allow Docker box to pull from a default registry

kokaz:
This allows user to pull images from a default Container Registry.

If we don't specify the registry but include it as a prefix in the repository (e.g. `wcr.io/wercker/my-repo`) we infer the registry URL and push to v2 by default.

Depends on #325


tjames221188:
Looks good to me

docker/box.go
```diff
@@ -543,14 +544,25 @@ func (b *DockerBox) Fetch(ctx context.Context, env *util.Environment) (*docker.I
...
+
+	b.config.Auth.Interpolate(env)
+
+	// If user use Azure or AWS container registry we don't infer.
+	if b.config.Auth.AzureClientSecret == "" && b.config.Auth.AwsSecretKey == "" {
+		repo, b.config.Auth = InferRegistry(repo, b.config.Auth, b.options)
+	}
+
+	if b.config.Auth.Registry == b.options.WerckerContainerRegistry.String() {
+		b.config.Auth.Username = "token"
```
jwolski2: Pull `"token"` out as a constant and leave a comment about it. Something like:

"The Username value is arbitrary. It is unused by the callees. Therefore, the value can be anything so long as it's not empty. "

kokaz: @jwolski2 Fixed

PR #335, opened 2017-08-31
make docker stuff easier to use 

tjames221188:
[#703](https://github.com/wercker/backlog/issues/703) 

- New `wercker docker ...` command - injects wercker credentials into docker config, then shells out to docker cli.
- tests around injecting config


cmd/docker.go
```diff
@@ -0,0 +1,46 @@
...
+	"os"
+	"os/exec"
+
+	"github.com/docker/cli/cli/config"
+	"github.com/docker/docker/api/types"
+	"github.com/kr/pty"
+	"github.com/wercker/wercker/core"
+)
+
+var errNoWerckerToken = fmt.Errorf("Wercker auth token could not be found, please run wercker login")
```
kokaz: please use `errors.New` instead.

`Errorf` is for formatting error which is not the case here.

cmd/main.go
```diff
@@ -322,6 +322,29 @@ var (
...
+			env := util.NewEnvironment(os.Environ()...)
+			opts, err := core.NewWerckerDockerOptions(settings, env)
+			if err != nil {
+				cliLogger.Errorln("Invalid options\n", err)
+				os.Exit(1)
+			}
+
+			err = ensureWerckerCredentials(opts)
+			if err != nil {
+				cliLogger.Errorln("Error ensuring wercker credentials: \n", err)
```
kokaz: There is a superfluous space after the colon character

cmd/docker.go
kokaz: Also error message should not start with a capital letter

cmd/docker.go
```diff
@@ -0,0 +1,46 @@
...
+	dockerConfig := config.LoadDefaultConfigFile(os.Stderr)
+	_, hasAuth := dockerConfig.AuthConfigs[opts.WerckerContainerRegistry.String()]
+	if !hasAuth {
+		dockerConfig.AuthConfigs[opts.WerckerContainerRegistry.String()] = types.AuthConfig{
+			Username: "token",
+			Password: opts.AuthToken,
+		}
+		err := dockerConfig.Save()
+		if err != nil {
+			return fmt.Errorf("Could not inject wercker token into docker config %v", err)
```
kokaz: Probably put a colon before `%v` like you did else where

cmd/docker_test.go
```diff
@@ -0,0 +1,90 @@
...
+	tempDir := s.WorkingDir()
+	config.SetDir(tempDir)
+	filename := path.Join(tempDir, "config.json")
+	data := []byte("{}")
+
+	err := ioutil.WriteFile(filename, data, 0644)
+	return err
+}
+
+func (s *DockerSuite) TestEnsureWerckerCredentialsNoToken() {
```
kokaz: We should only test exported functions. That being said you have two choices:

1. Export `ensureWerckerCredentials`
2. Test against the `func` in `dockerCommand.Action` (https://github.com/wercker/wercker/pull/335/files#diff-3002d695552ec7e7ebbaaf04a21d3439R330)

I think (1) is enough

hatchan: I see no reason why we cannot test unexported functions

PR #340, opened 2017-09-27
Preserve timestamp when copying source to container

hatchan:
This should preserve the timestamp of the git repository or the local files. A user was having a specific issue which had to do with golang rebuilding every package every time. Even though the user had every file in the cache (`go install` uses timestamps to determine if a package needs to be rebuild)

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -2,6 +2,7 @@
 - Update dependencies, as a result of `Sirupsen/logrus` -> `sirupsen/logrus` (#333)
 - Add a Docker subcommand (#335)
 - Ensure repository names are always lowercase (#338)
+- Preservce timestamps when copying source and cache into container (#340)
 
 ## v1.0.965 (2017-08-23)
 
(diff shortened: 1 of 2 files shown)


changelog.md
```diff
@@ -2,6 +2,7 @@
 - Update dependencies, as a result of `Sirupsen/logrus` -> `sirupsen/logrus` (#333)
 - Add a Docker subcommand (#335)
 - Ensure repository names are always lowercase (#338)
+- Preservce timestamps when copying source and cache into container (#340)
```
jwolski2: Typo `Preservce` -> `Preserve`

hatchan:
@jwolski2 sorry updated the ticket with a description.

hatchan:
After testing this it turns out that it doesn't work yet.

garthy:
I belive this doesn't work because the project copy in "func (p *Runner) EnsureCode() (string, error)" doesn't use "cp" and uses shutil.CopyTree which doesn't preserve timestamps.

PR #342, opened 2017-10-09
Interpolate the box name when starting service

leesio:
We'd like to be able to tag images with their commit SHA, push to a
container registry, pull down this same image in a subsequent
pipeline and run it as a service. In order to do that, we need the `tag`
field of the service configuration to be interpolated. 

Are there any issues with this implementation?

As far as I can tell it should have no negative impact.

### DIFF
--- a/docker/service.go
+++ b/docker/service.go
@@ -185,7 +185,7 @@ func (b *InternalServiceBox) Run(ctx context.Context, env *util.Environment, lin
 	}
 
 	conf := &docker.Config{
-		Image:           b.Name,
+		Image:           env.Interpolate(b.Name),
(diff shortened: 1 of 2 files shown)


jwolski2:
@leesio Agreed that this _should_ work. Can you add a test to verify the container's configuration is as expected (for non-substituted and substituted image names)?

leesio:
@jwolski2 I could write a test for that. It seems there is 0% coverage for the `service` component at the moment though 🙁

leesio:
@termie thanks for looking into this and thanks for the suggestion. I'll look into adding a relevant test project.

leesio:
Is there any update on this?

PR #359, opened 2018-03-06
Fixes #357 - Check for errors in starting box container

### LINKED ISSUE #357: Wercker cli on mac will fail due to mount issues
Running wercker build on mac will fail when the directories on Host being mounted to the box ( such as the directory specified by option --working-dir) are not explicitly "shared" in Docker-For-Mac preferences screen

file_sharing_screen

Also while starting box container, we do not check if any error was returned while starting the container and therefore we get errors in later stage:

--> Executing pipeline
ERROR --> Step failed: setup environment 9.61s
Guest command failed with exit code -1: mkdir -p "/pipeline"
ERROR Guest command failed with exit code -1: mkdir -p "/pipeline"
WARNING Box container has already stopped.
FATAL Exiting.

because the box has not started yet. Instead we should get an error immediately after the box have not started - like

--> Executing pipeline
ERROR --> Step failed: setup environment 15.81s
API error (502): Mounts denied:

lake-of-dreams:
Fixes [357](https://github.com/wercker/wercker/issues/357) - Check for error after starting box. There was an extraneous check on error being null, moved the code to start container above it and captured the error
[Wercker Build run url](https://app.wercker.com/wercker/wercker/runs/build/5a9e9a5f88122900019f1ca9)

### DIFF
--- a/docker/box.go
+++ b/docker/box.go
@@ -438,11 +438,12 @@ func (b *DockerBox) Run(ctx context.Context, env *util.Environment) (*docker.Con
 
 	b.logger.Debugln("Docker Container:", container.ID)
 
+	err = client.StartContainer(container.ID, hostConfig)
+
 	if err != nil {
 		return nil, err
 	}
 
-	client.StartContainer(container.ID, hostConfig)
 	b.container = container
 	return container, nil
 }


docker/box.go
```diff
@@ -438,11 +438,12 @@ func (b *DockerBox) Run(ctx context.Context, env *util.Environment) (*docker.Con
 
 	b.logger.Debugln("Docker Container:", container.ID)
 
+	err = client.StartContainer(container.ID, hostConfig)
+
```
hatchan: Can you remove this line.

lake-of-dreams: Oops - sorry missed this comment before merging - I kept the line because in code above, b/w create container and check for error - there was a line - so just left it for better readability. If you  will suggest - I can fix in a different branch and raise PR please?

hatchan: I always do error handling immediately after the code that captures it. And then have a new line after that. So you get little blocks of related code. You can create a new PR with the fix, but it is not that big a deal.

lake-of-dreams: Sure. Raised PR https://github.com/wercker/wercker/pull/360 - Please have a look

PR #361, opened 2018-03-07
Fix for #1650:  In session.go, SendChecked sometimes hangs #1650 

nigeldeakin:
See https://github.com/wercker/backlog/issues/1650 for a description of the issue which this PR fixes.

Essentially, this method previously only handled the case where the sentinel at the start of the line.

This fix handles the case (which I have seen intermittently, about 1 in 5 times when running in the CLI) when the sentinel string is appended to some previous output. When this occurs the sentinel string (and any characters that follow it - see note 1 below) are moved to a separate line.

Note 1:  Why do I copy the sentinel string *and any characters that follow it* to a separate line? The reason is that the the sentinel appears to be always followed by a space and an integer, 0 in the cases I have seen.  This is explicitly tested in the `checkLine` function, which is used by the caller to test whether a line returned by this function contains a sentinel.

### DIFF
--- a/core/session.go
+++ b/core/session.go
@@ -199,19 +199,26 @@ func smartSplitLines(line, sentinel string) []string {
 		possibleSentinel := splitLines[len(splitLines)-2]
 		// And we expect a newline at the end
 		possibleSentinel = fmt.Sprintf("%s\n", possibleSentinel)
-		foundExit, _ := checkLine(possibleSentinel, sentinel)
-		// If we found the exit code, make sure it gets read as a separate line
-		if foundExit {
+
+		// does this string contain the sentinel?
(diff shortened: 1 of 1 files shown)


hatchan:
The trailing integer is the exit code.

nigeldeakin:
`checkLine` (which has not changed) checks whether a specified line starts with the sentinel, followed by an integer (which I know know is an exit code).

After this change, `checkLine` is called from just one place: the anonymous function in `SendChecked`, where it is used to check whether one of the lines returned by `splitLines` starts with the sentinel and exit code.

Following my proposed change to `splitLines`,  the sentinel will always occur at the start of a line, which means that `checkLine` will now work correctly.

PR #362, opened 2018-03-09
New docker-build step and modified docker-push

nigeldeakin:
This PR introduces a new internal step `internal/docker-build` to build an image from a Dockerfile, and extends the existing internal step `internal/docker-push` so that it can push this image to a registry.

The latest specification of this feature is [Dockerfile support (v3)](https://wercker.quip.com/exSMAVWrScbV)

An example of the use of these new steps is at https://github.com/nigeldeakin/docker-build-golang . It is intended that this be moved to be under [the wercker organisation](https://github.com/wercker) in due course.

More testing is needed before this PR can be merged, including the addition of new automated test cases.

For `docker-build`, the only properties supported are `dockerfile` and `tag`. Other properties will be added later. The specification will be updated to reflect this.

For `docker-push`, the existing behaviour for creating, tagging and pushing an image from the running pipeline is cchanged. A new property `id` has been added. This is used to specify that an existing image (i.e. one created using `docker-build`) should be tagged and pushed. When used in this mode, the properties that specify what tag should be added, what registry should be used, and to supply  authentication information, will be supported, whereas any properties which relate to committing a new image from the current pipeline container will be ignored. 

Other outstanding issues with `docker-build`:
* Not all properties are supported yet. Only `dockerfile` and `tag`.
* The image that is built will remain in the local docker daemon. We probably want this to be deleted automatically at the end of the pipeline, with a user option to disable this in the CLI.
* Images created by intermediate build stages, and images created by intermediate layers within a build stage, will remain. We probably want to these to be deleted automatically at the end of a step, with a user option to disable this in the CLI.
* Note that intermediate containers that are created during the execution of this dockerfile will be deleted automatically unless the user specifies `-rm` (though this flag is not yet implemented).
* There is a minor UI infelicity in that when the docker build downloads an image, the animated arrow that is seen when a `docker build` is run directly is shown as multiple lines, making the output more verbose. 

Outstanding issues with `docker-push`
* Need to fail the step is the the user specifies the property`image` at the same time as specifying a property which is not appropriate when pushing an existing image. This has not yet been implemented. Currently any inappropriate properties will be simply ignored.


nigeldeakin:
I've now added support for the properties `buildargs` and `q` (quiet).

I have also added code to parse the properties `extrahosts`, `squash` and labels` but these settings are not actually used as the Docker golang client does not supported these properties. 

I'm aware that the parsing of `buildargs` has a bug https://github.com/wercker/backlog/issues/1676 which means that it doesn't honour commas as a delimiter between key=value pairs. I have not attempted to fix this as the same bug applies in other places and we need a consistent fix for all places that allow key-value pairs to be specified.

docker/docker-build.go
```diff
@@ -0,0 +1,382 @@
...
+		}
+	}
+
+	if qProp, ok := s.data["q"]; ok {
+		q, err := strconv.ParseBool(qProp)
+		if err == nil {
+			s.q = q
+		} else {
+			// bad value, default to false (verbose)
+			s.q = true
```
hatchan: You can move this to line 121, then remove this `else` statement and the `else` statement on line 130.

docker/docker-build.go
```diff
@@ -0,0 +1,382 @@
...
+		}
+	}
+
+	if squashProp, ok := s.data["squash"]; ok {
+		squash, err := strconv.ParseBool(squashProp)
+		if err == nil {
+			s.squash = squash
+		} else {
+			// bad value, default to false (do not squash)
+			s.squash = false
```
hatchan: Same as the the q parameter.

docker/docker-build.go
```diff
@@ -0,0 +1,382 @@
...
+
+// Fetch NOP
+func (s *DockerBuildStep) Fetch() (string, error) {
+	// nop
+	return "", nil
+}
+
+// Execute builds an image
+func (s *DockerBuildStep) Execute(ctx context.Context, sess *core.Session) (int, error) {
+
```
hatchan: This can go

docker/docker-build.go
```diff
@@ -0,0 +1,382 @@
...
+	s.logger.Debugln("Building input tar", s.options.HostPath(destTar))
+
+	layerFile, err := os.OpenFile(s.options.HostPath(destTar), os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0644)
+	if err != nil {
+		return err
+	}
+	defer layerFile.Close()
+
+	digester := digest.Canonical.Digester()
```
hatchan: This does not appear to be used, do you require it?

docker/docker.go
```diff
@@ -942,27 +950,36 @@ func (s *DockerPushStep) Execute(ctx context.Context, sess *core.Session) (int,
...
-	return s.tagAndPush(i.ID, e, client)
+		if s.dockerOptions.CleanupImage {
+			defer cleanupImage(s.logger, client, s.repository, s.tags[0])
+		}
+
+		s.logger.WithField("Image", i).Debug("Commit completed")
+		imageID = i.ID
+	} else {
+		// id is set to the name or ID of an existing image which will be tagged and pushed (equivalent to "docker push")
+		imageID = s.image
```
hatchan: This can be set to the initial value on line 953.

docker/docker-build.go
```diff
@@ -0,0 +1,325 @@
...
+		BaseStep:      baseStep,
+		data:          stepConfig.Data,
+		logger:        util.RootLogger().WithField("Logger", "DockerBuildStep"),
+		options:       options,
+		dockerOptions: dockerOptions,
+	}, nil
+}
+
+func (s *DockerBuildStep) configure(env *util.Environment) {
+
```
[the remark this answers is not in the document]
nigeldeakin: You mean never have a blank line at the start of a function?
OK

docker/docker-build.go
nigeldeakin: There's a more serious error here, in that the default should be `false`, not `true`. (Docker build is verbose by default). I have changed that.

I've removed the `else` statements as requested. I included them because I like my code to be explicit, for example to show that if the value is not a boolean then it the error is deliberately ignored and the default value used. Without this a reader might think I have simply forgotten to handle this case, and the current behaviour is an oversight. I have added a comment to explain this.

docker/docker-build.go
nigeldeakin: Done

docker/docker-build.go
nigeldeakin: Done

docker/docker.go
nigeldeakin: I like having two separate blocks here as it reflects the two different modes in which `internal/docker-push` can be used: the classic mode, when you're committing and pushing the pipeline container, and the new mode when you're pushing an image that you created previously. I feel we're sacrificing readability by getting rid of the `else` case and setting a default which we overwrite in the `if` block.

docker/docker-build.go
nigeldeakin: This is the code in question:
```
digester := digest.Canonical.Digester()
mwriter := io.MultiWriter(layerFile, digester.Hash())
tw := tar.NewWriter(mwriter)
```
This was copied from `docker.go`, function `(s *DockerScratchPushStep) Execute`, which did the same. In addition to writing files to the tar file, it's also writing them to the running hash of the `Digester`. However the digester is not used for anything so appears to be redundant. I have changed this to
```
tw := tar.NewWriter(layerFile)
```

hatchan: Yeah saw it was used on the next line, but nothing besides that. :+1: on the changes.

nigeldeakin:
I've added an integration test for the `internal/docker-build` step

nigeldeakin:
I've made some changes: 

`docker-build` no longer sets the tag on the new image directly. Instead it sets a `image-name` which is `docker-build` prepends with build ID and uses to set the image's repository and tag. 

`docker-push` has been changed correspondingly. The `image` property has been renamed `image-name`. `docker-push` will prepend this with the build ID when working out which image to push.

Tests updated.

PR #366, opened 2018-03-16
Internal docker run

snanurag:
This PR introduces a new internal step internal/docker-run to run a container from an image.

The latest specification of this feature is [Dockerfile support (v4)](https://wercker.quip.com/ysCpAUenLGiC).

More testing and more of Unit and Functional tests are needed before this PR can be merged.


docker/docker_run.go
```diff
@@ -0,0 +1,219 @@
...
+}
+
+// InitEnv parses our data into our config
+func (s *DockerRunStep) InitEnv(hostEnv *util.Environment) {
+	env := s.Env()
+	s.configure(env)
+}
+
+func (s *DockerRunStep) configure(env *util.Environment) {
+
```
hatchan: Can you remove this empty line

docker/docker_run.go
```diff
@@ -0,0 +1,219 @@
...
+// Fetch NOP
+func (s *DockerRunStep) Fetch() (string, error) {
+	// nop
+	return "", nil
+}
+
+// Execute creates the container and starts the container.
+func (s *DockerRunStep) Execute(ctx context.Context, sess *core.Session) (int, error) {
+
```
hatchan: Can you remove this line

docker/docker_run.go
```diff
@@ -0,0 +1,219 @@
...
+		dockerOptions: dockerOptions,
+	}, nil
+}
+
+// InitEnv parses our data into our config
+func (s *DockerRunStep) InitEnv(hostEnv *util.Environment) {
+	env := s.Env()
+	s.configure(env)
+}
+
```
nigeldeakin: That doesn't look right. Why are we ignoring the input parameter?

docker/docker_run.go
```diff
@@ -0,0 +1,219 @@
...
+	hostconfig := &docker.HostConfig{
+		DNS:          s.dockerOptions.DNS,
+		PortBindings: s.portBindings,
+		Links:        s.links,
+	}
+
+	s.createContainer(client, conf, hostconfig)
+
+	s.startContainer(client, hostconfig)
+
```
nigeldeakin: Both these functions return an error code. Is there a reason why you are ignoring these?

This could be a good place to log that the container was successfully started - or failed.

docker/docker_run.go
```diff
@@ -0,0 +1,219 @@
...
+}
+
+func (s *DockerRunStep) configure(env *util.Environment) {
+
+	if ports, ok := s.data["ports"]; ok {
+		parts, err := shlex.Split(ports)
+		if err == nil {
+			s.portBindings = portBindings(parts)
+			s.exposedPorts = exposedPorts(parts)
+		}
```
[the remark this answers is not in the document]
snanurag: Well, my take on this is - Ports binding are publishing (exposedPorts) are generic features of containers. This is not specific to box container or service or docker-run container. These both functions are defined in Box.go. Box, service and docker-run all, are referring same functions. So, if there are any chances of anything wrong, those apply to all containers that Wercker starts. And, we would require to address them at bigger level than scope of this task.

docker/docker_run.go
snanurag: fixed.

docker/step.go
```diff
@@ -38,6 +38,9 @@ func NewStep(config *core.StepConfig, options *core.PipelineOptions, dockerOptio
 	if config.ID == "internal/publish-step" {
 		return NewPublishStep(config, options, dockerOptions)
 	}
```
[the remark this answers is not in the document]
snanurag: fixed

docker/docker_run.go
snanurag: fixed

docker/docker_run.go
snanurag: fixed

docker/docker_run.go
```diff
@@ -0,0 +1,219 @@
+//   Copyright @2018, Oracle and/or its affiliates. All rights reserved.
```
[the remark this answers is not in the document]
snanurag: fixed

docker/docker_run_test.go
```diff
@@ -0,0 +1,89 @@
+//   Copyright @2018, Oracle and/or its affiliates. All rights reserved.
```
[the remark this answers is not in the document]
snanurag: fixed

docker/docker_run.go
snanurag: fixed

docker/docker_run.go
```diff
@@ -0,0 +1,302 @@
...
+		data:                  stepConfig.Data,
+		logger:                util.RootLogger().WithField("Logger", "DockerRunStep"),
+		options:               options,
+		dockerOptions:         dockerOptions,
+		OriginalContainerName: originalContainerName,
+	}, nil
+}
+
+// InitEnv parses our data into our config
+func (s *DockerRunStep) InitEnv(env *util.Environment) {
```
lake-of-dreams: If you will merge with master now - you will need to return an error from here..

docker/docker_run.go
```diff
@@ -0,0 +1,302 @@
...
+	if user, ok := s.data["user"]; ok {
+		s.User = env.Interpolate(user)
+	}
+}
+
+func getCorrectImageName(env *util.Environment, s *DockerRunStep) string {
+	i := env.Interpolate(s.data["image"])
+
+	client, err := NewDockerClient(s.dockerOptions)
+	if err != nil {
```
lake-of-dreams: this seems like a fatal error - should be returned and propagated upwards

docker/docker_run.go
```diff
@@ -0,0 +1,302 @@
...
+}
+
+// Fetch NOP
+func (s *DockerRunStep) Fetch() (string, error) {
+	return "", nil
+}
+
+// Execute creates the container and starts the container.
+func (s *DockerRunStep) Execute(ctx context.Context, sess *core.Session) (int, error) {
+	networkName := s.dockerOptions.NetworkName
```
lake-of-dreams: How does one pass an networkName to this step?

lake-of-dreams: And you should check if this network indeed exist or not

docker/docker_run.go
```diff
@@ -0,0 +1,302 @@
...
+	EntryPoint            []string
+	WorkingDir            string
+	PortBindings          map[docker.Port][]docker.PortBinding
+	ExposedPorts          map[docker.Port]struct{}
+	User                  string
+	ContainerName         string
+	OriginalContainerName string
+	Image                 string
+	ContainerID           string
+	Repository            string
```
lake-of-dreams: What is the difference b/w image and Repository here?

docker/docker_run.go
```diff
@@ -0,0 +1,302 @@
...
+		}
+	}
+
+	if user, ok := s.data["user"]; ok {
+		s.User = env.Interpolate(user)
+	}
+}
+
+func getCorrectImageName(env *util.Environment, s *DockerRunStep) string {
+	i := env.Interpolate(s.data["image"])
```
lake-of-dreams: what happens when there is no "image" in s.data?

lake-of-dreams: sone doc on this will help explain what it exactly intends to do

docker/docker_run.go
```diff
@@ -0,0 +1,302 @@
...
+
+	client, err := NewDockerClient(s.dockerOptions)
+	if err != nil {
+		return ""
+	}
+
+	// local image should exists with a prepend of build id.
+	image, err := client.InspectImage(s.options.RunID + i)
```
lake-of-dreams: so essentially docker-run will allow not only running the images built locally but also any image one might provide using the "image" field - correct?

docker/docker_run.go
snanurag: I changed some code. Now network name is retrieved from Box only. Network existence is checked in Box before starting the box container. If it hadn't existed then pipeline would have failed even before step execution starts. And, after the box is started on a network, that network can't be deleted without clearing box container.

docker/docker_run.go
snanurag: repository is not needed now. I forgot to clear up the code. Donw now.

docker/docker_run.go
```diff
@@ -0,0 +1,302 @@
...
+
+// Execute creates the container and starts the container.
+func (s *DockerRunStep) Execute(ctx context.Context, sess *core.Session) (int, error) {
+	networkName := s.dockerOptions.NetworkName
+	if networkName == "" {
+		networkName = s.options.RunID
+	}
+
+	boxConfig := &core.BoxConfig{
+		ID: s.Image,
```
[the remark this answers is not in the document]
snanurag: image could be whatever we provide in `docker pull` or as box `id`. It is in consistency with box and service `id`. So, it could be golang or golang:latest or docker.io/library/golang:latest or wcr.io/library/xyz:v1.

docker/docker_run.go
snanurag: made changes to handle error more appropriately. And, we keep updating this in doc here - confluence.oraclecorp.com/confluence/display/WRKR/Dockerfile+and+docker+support

docker/docker_run.go
snanurag: Yes... that was the understanding before we started building it. I have provided more checks now.

docker/docker_run.go
lake-of-dreams: sounds good

docker/docker_run.go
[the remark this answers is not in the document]
lake-of-dreams: ok - make sure the documentation reflects that - because for such cases username and password will also be required to be provided with docker-run step configuration (Auth struct) while for "built" images - you will not need it.

docker/docker_run.go
lake-of-dreams: by doc I meant comments on method - someone changing this code in future might not have a handle to this confluence link

docker/docker_run.go
```diff
@@ -0,0 +1,319 @@
...
+	Image                 string
+	ContainerID           string
+	Auth                  dockerauth.CheckAccessOptions `yaml:",inline"`
+}
+
+type BoxDockerRun struct {
+	*DockerBox
+}
+
+// NewDockerRunStep is a special step for doing docker runs
```
lake-of-dreams: I think comment should elaborate a bit what it does exactly.. different between how this be used with an image built withing pipeline and any other image ( usage of "image" attribute)

tests/projects/docker-run/wercker.yml
```diff
@@ -0,0 +1,22 @@
+build:
+  box: golang
+  steps:
+    - internal/docker-build:
+        image-name: docker-push-image-name-1
+    - internal/docker-run:
+        image: docker-push-image-name-1
+        name: myTestContainer
+    - script:
+        name: Test Container
```
agnihotrigaurav: Is it possible to provide name like ${Test container} If yes then NewDockerRunStep requires modification while storing name.
If no, then Interpolate call "s.ContainerName = s.options.RunID + env.Interpolate(s.OriginalContainerName)" in configure method is not required.

tests/projects/docker-run/wercker.yml
```diff
@@ -0,0 +1,22 @@
+build:
+  box: golang
+  steps:
+    - internal/docker-build:
+        image-name: docker-push-image-name-1
+    - internal/docker-run:
+        image: docker-push-image-name-1
+        name: myTestContainer
```
agnihotrigaurav: Is it possible to provide name like ${Test container} If yes then NewDockerRunStep requires modification while storing name.
If no, then Interpolate call "s.ContainerName = s.options.RunID + env.Interpolate(s.OriginalContainerName)" in configure method is not required.

docker/docker_run.go
```diff
@@ -0,0 +1,319 @@
...
+
+func (s *DockerRunStep) configure(env *util.Environment) error {
+	if s.options.ExposePorts {
+		if ports, ok := s.data["ports"]; ok {
+			parts, err := shlex.Split(ports)
+			if err == nil {
+				s.PortBindings = portBindings(parts)
+				s.ExposedPorts = exposedPorts(parts)
+			} else {
+				return err
```
agnihotrigaurav: Log error before returning.

docker/docker_run.go
```diff
@@ -0,0 +1,319 @@
...
+		}
+	}
+
+	if workingDir, ok := s.data["working-dir"]; ok {
+		s.WorkingDir = env.Interpolate(workingDir)
+	}
+
+	image, err := getCorrectImageName(env, s)
+	if err != nil {
+		return err
```
agnihotrigaurav: Log error before returning

docker/docker_run.go
```diff
@@ -0,0 +1,319 @@
...
+	}
+
+	s.ContainerName = s.options.RunID + env.Interpolate(s.OriginalContainerName)
+
+	if cmd, ok := s.data["cmd"]; ok {
+		parts, err := shlex.Split(cmd)
+		if err == nil {
+			s.Cmd = parts
+		} else {
+			return err
```
agnihotrigaurav: Log error before returning

docker/docker_run.go
```diff
@@ -0,0 +1,319 @@
...
+		}
+	}
+
+	if entryPoint, ok := s.data["entrypoint"]; ok {
+		parts, err := shlex.Split(entryPoint)
+		if err == nil {
+			s.EntryPoint = parts
+		} else {
+			return err
```
agnihotrigaurav: Log error before returning

docker/docker_run.go
```diff
@@ -0,0 +1,319 @@
...
+		parsedEnv, err := shlex.Split(envi)
+
+		if err == nil {
+			interpolatedEnv := make([]string, len(parsedEnv))
+			for i, envVar := range parsedEnv {
+				interpolatedEnv[i] = env.Interpolate(envVar)
+			}
+			s.env = interpolatedEnv
+		} else {
+			return err
```
agnihotrigaurav: Log error before returning

docker/docker_run.go
```diff
@@ -0,0 +1,319 @@
...
+	}
+
+	if user, ok := s.data["user"]; ok {
+		s.User = env.Interpolate(user)
+	}
+	return nil
+}
+
+func getCorrectImageName(env *util.Environment, s *DockerRunStep) (string, error) {
+	i := env.Interpolate(s.data["image"])
```
agnihotrigaurav: Kindly log error before or after returning getCorrectImageName function

docker/docker_run.go
snanurag: I don't think these errors required logging. `Init` function definition of `Step` interface is changed recently. It further throws the errors. So it depends on the framework of `Step` interface - whether it wants to log those errors or display them on the console. So, whatever errors are thrown from `configure` method are related to properties input. They are on display to the users. They are not developer errors, so not required logging

tests/projects/docker-run/wercker.yml
snanurag: I have removed env.Interpolate as of now. There was no requirement as such to interpret environment variables in name. If such requirement will come up then I will change.

docker/docker_run.go
snanurag: I realized that the more checks I had provided were wrong. I started checking the existence of input image if `docker-build` image doesn't exist. It was supposed to pull the input image if `docker-build` image doesn't exist. Fixed it now. And, comments are provided as well.

tests/projects/docker-run/wercker.yml
nigeldeakin: Are you sure? There seems to be a general convention that you can use environment variables in properties, I notice. For example the docs for `docker-push` explicitly says "It is possible to use environment variables inside all properties, which will be expanded. "It is possible to use environment variables inside all properties, which will be expanded."   (http://devcenter.wercker.com/docs/steps/internal-steps) My preference would be to be consistent and allow env vars here.

snanurag: I have supported it all places now except `entrypoint`, `cmd` and `ports`. I have not supported it in these 3 properties to be consistent with service.

PR #371, opened 2018-04-02
Fix #1741 Correct status reporting from docker push by parsing status messages from docker

lake-of-dreams:
Fixes [1741](https://github.com/wercker/backlog/issues/1741) 
Both internal/docker-scratch-push and internal-push steps invoke tagAndPush method which further invokes DockerClient.PushImage().In a situation where docker push call will fail due to any reason - one of which being incorrect username/password for the repository, the error is not correctly captured as the REST api call for push to docker daemon immediately returns without any error. The error message is in-fact present in progress status messages from the daemon but we were not looking at those status messages to check whether a push was indeed successful. In case of a successful push, the tag, size and SHA digest of image being pushed is present in status progress messages. In this PR - logic has been changed to not just rely on the error returned by docker client, but look for the actual successful status message in streamed statuses.

**Summary of changes**

1. in docker/docker.go - parse status messages from docker daemon for a push call. Relevant structs were created to correctly marshal the message ( The current client we use does not have structs to marshal these messages directly - We should probably refactor this code once we upgrade to official docker client in [1698](https://github.com/wercker/backlog/issues/1698) - a comment has been added to reflect this.
2. Added unit tests to docker/docker_push_test.go
3. Added -v to govendor test in wercker.yml to be able to see what tests are being run during the build
4. Moved call to cleanupImage a bit up

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,5 +1,10 @@
 ## unreleased
 
+## v1.0.XXXX(2018-04-XX)
+
+- Fix status reporting for docker push (#371)
+
+
 ## v1.0.1183 (2018-03-29)
 
 - New docker-build step and enhanded docker-push step (#362)
--- a/changelog.md
+++ b/changelog.md
@@ -1,9 +1,6 @@
 ## unreleased
 
-## v1.0.XXXX(2018-04-XX)
-
 - Fix status reporting for docker push (#371)
-
 
 ## v1.0.1183 (2018-03-29)
 
--- a/docker/docker.go
+++ b/docker/docker.go
@@ -60,7 +60,7 @@ const (
 
 //TODO: The current fsouza/go-dockerclient does not contain structs for status messages emitted
 // from docker in case of push - therefore had to explicitly create these structs for better
-// usablity of code (instead of decofing json to a map). Official docker client should contain
+// usablity of code (instead of unmarshalling json to a map). Official docker client should contain
 // these structs(or equivalents) already and this code should be refactored to use those instead
 // having to maintain our own.
 
--- a/docker/docker.go
+++ b/docker/docker.go
@@ -16,9 +16,11 @@ package dockerlocal
 
 import (
 	"archive/tar"
+	"bytes"
 	"crypto/rand"
 	"crypto/sha256"
 	"encoding/hex"
+	"encoding/json"
 	"fmt"
 	"io"
 	"net/url"
@@ -40,6 +42,7 @@ import (
 	"github.com/google/shlex"
 	digest "github.com/opencontainers/go-digest"
 	"github.com/pborman/uuid"
+	"github.com/pkg/errors"
 	"github.com/wercker/docker-check-access"
 	"github.com/wercker/wercker/auth"
 	"github.com/wercker/wercker/core"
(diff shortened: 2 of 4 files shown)


lake-of-dreams:
Have been tested using canary experiment on staging

docker/docker.go
```diff
@@ -1024,11 +1071,36 @@ func (s *DockerPushStep) tagAndPush(imageID string, e *core.NormalizedEmitter, c
...
+				s.logger.Errorln("Failed to parse status outputs from docker push:", err)
+				return 1, err
+			}
+			isContainerPushed := false
+			for _, statusMessage := range statusMessages {
+				if len(strings.TrimSpace(statusMessage.Error)) != 0 {
+					errorMessageToDisplay := statusMessage.Error
+					if statusMessage.ErrorDetail != nil {
+						jsonMessage, _ := json.Marshal(statusMessage.ErrorDetail)
+						errorMessageToDisplay = string(jsonMessage)
```
hatchan: Why convert it back to a JSON byte array and then back to a string? Can't you use `statusMessage.ErrorDetail.Message` directly?

docker/docker.go
```diff
@@ -1024,11 +1071,36 @@ func (s *DockerPushStep) tagAndPush(imageID string, e *core.NormalizedEmitter, c
...
+			statusJSON := strings.Join([]string{"[", strings.TrimSuffix(strings.Replace(buf.String(), "\n", ",", -1), ","), "]"}, "")
+			statusMessages := make([]PushStatus, 0)
+			err = json.Unmarshal(([]byte)(statusJSON), &statusMessages)
+			if err != nil {
+				s.logger.Errorln("Failed to parse status outputs from docker push:", err)
+				return 1, err
+			}
+			isContainerPushed := false
+			for _, statusMessage := range statusMessages {
+				if len(strings.TrimSpace(statusMessage.Error)) != 0 {
```
hatchan: How certain are you that this is a fatal error? Ie. isn't it possible that Docker retries or that the error can be ignored?

docker/docker.go
```diff
@@ -52,8 +55,45 @@ const (
 	// so the value can be anything so long as it's not empty.
 	DefaultDockerRegistryUsername = "token"
 	DefaultDockerCommand          = `/bin/sh -c "if [ -e /bin/bash ]; then /bin/bash; else /bin/sh; fi"`
+	NoPushConfirmationInStatus    = "Docker push failed to complete.Please check logs for any error condition"
```
hatchan: Add a space after the first `.`, and add a trailing `.`.

docker/docker.go
```diff
@@ -52,8 +55,45 @@ const (
...
 
+//TODO: The current fsouza/go-dockerclient does not contain structs for status messages emitted
+// from docker in case of push - therefore had to explicitly create these structs for better
+// usablity of code (instead of unmarshalling json to a map). Official docker client should contain
+// these structs(or equivalents) already and this code should be refactored to use those instead
+// having to maintain our own.
+
+//PushStatusAux : The "aux" component of status message
+type PushStatusAux struct {
+	Tag    string `json:"Tag,omitempty"`
```
hatchan: Are you sure the keys are with capital letters? None of the other structs in this file have that.

changelog.md
```diff
@@ -1,5 +1,10 @@
 ## unreleased
 
+## v1.0.XXXX(2018-04-XX)
```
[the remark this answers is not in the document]
lake-of-dreams: Done

docker/docker.go
lake-of-dreams: Done

docker/docker.go
lake-of-dreams: Fixed. It worked this was as well because unmarshalling seems to work on non-caps keys as well https://github.com/golang/go/issues/11673

docker/docker.go
lake-of-dreams: statusMessage.ErrorDetail also contains a "code" - so to capture that as well in the error message being displayed and since we do not have any standard to display error messages with error codes - I thought it will be better to display it as a json message and instead of creating a json message myself from statusMessage.ErrorDetail .Message and statusMessage.ErrorDetail.Code - I unmarshal it to json

docker/docker.go
```diff
@@ -1024,11 +1071,36 @@ func (s *DockerPushStep) tagAndPush(imageID string, e *core.NormalizedEmitter, c
 				s.logger.Errorln("Failed to push:", err)
 				return 1, err
 			}
-			s.logger.Println("Pushed container:", s.repository, s.tags)
+			// Covert status messages in stream {...} {...} to a proper json array [{...},{...},...]
+			statusJSON := strings.Join([]string{"[", strings.TrimSuffix(strings.Replace(buf.String(), "\n", ",", -1), ","), "]"}, "")
```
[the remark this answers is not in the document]
lake-of-dreams: Fixed

docker/docker.go
lake-of-dreams: These status messages are from docker push and this particular error was if we are unable to decode the json message coming out of docker - these messages are after docker have done all retries - so in a way was fatal because if we are unable to decode a message - in forthcoming lines we will not be able to identify if a push was successful. But since I check for presence of sha anyway - I have removed the "fatality" of this decode error but not returning from here in a error condition

docker/docker.go
hatchan: This is being displayed in our interface (cli or web) isn't it? So why would you want JSON there? You could have easily done a `fmt.Sprintf("Code: %s, Message: %s", ....)`

lake-of-dreams: ok I will fix this in next PR I will do for 1740

PR #372, opened 2018-04-03
Update docker-build to use official docker client and support more build parameters

nigeldeakin:
This PR updates the `internal/docker-build` step to use the official Docker client in order to allow some more build parameters to be supported.

This fixes https://github.com/wercker/backlog/issues/1819, which reported that `--no-remove` (in a local build) did not prevent intermediate containers from being removed.

This adds support for the following build properties:
* `labels`
* `extrahosts`
* `squash`

Still to do:
* Confirm that we really do need to support `extrahosts` and `labels` 
* Automated tests for all new properties


wercker.yml
```diff
@@ -1,20 +1,20 @@
-box: golang:1.7
+box: golang:1.9.4
```
hatchan: We can simply use the tag `1.9` here.

wercker.yml
```diff
@@ -1,20 +1,20 @@
...
+        packages: openssh-client pkg-config libsystemd-dev
 
     - script:
         name: install govendor
         code: go get -u github.com/kardianos/govendor
 
     - script:
-        name: go vet
-        code: govendor vet +local
+        name: go vet -composites=false
```
hatchan: This is fixed in https://github.com/wercker/wercker/pull/374 so probably best to wait till that is merged

docker/docker_build.go
```diff
@@ -123,6 +123,14 @@ func (s *DockerBuildStep) configure(env *util.Environment) {
 		}
 	}
 
+	s.nocache = false // default to false when value is bad or not set
+	if nocacheProp, ok := s.data["nocache"]; ok {
```
hatchan: Why aren't we using dashes between the words? Not sure if we ever had a discussion about that though. To me, having a dash in between there makes it more readable, plus it will match the original flag. Same goes for `extrahosts` below. Not sure if any existing flags are missing these dashes though...

docker/status_handler.go
```diff
@@ -42,6 +42,10 @@ type JSONMessageProcessor struct {
 // ProcessJSONMessage will take JSONMessage m and generate logs based on the
 // message and previous messages.
 func (s *JSONMessageProcessor) ProcessJSONMessage(m *jsonmessage.JSONMessage) string {
+	if m.Stream != "" {
```
hatchan: What is in here, and why are we ignoring the normal flow when it does contain something?

docker/docker_client_fsouza.go
```diff
@@ -0,0 +1,227 @@
+//   Copyright © 2016,2018, Oracle and/or its affiliates.  All rights reserved.
```
hatchan: I'm guessing no code has been changed below...

nigeldeakin:
Looks like the `extrahosts` property doesn't work with the version of the docker daemon that is being used in `dev.wercker.com`. (The daemon supports API version 1.24 but `extrahosts` requires API version 1.28) However we can leave in the implementation but simply remove it temporarily from the documentation. (Note that we will have to update the API verison used by the client before this feature will work)

wercker.yml
nigeldeakin: OK, reverted,

docker/docker_build.go
nigeldeakin: I was using the names as documented in the docker API at https://docs.docker.com/engine/api/v1.31/#operation/ImageList . But yes, I agree that matching the CLI flag is better. I'll make the necessary changes.

nigeldeakin: `nocache` changed to `no-cache`
`buildargs` changed to `build-args` (though note that the CLI equivalent is to have multiple `--build-arg` parameters rather than supplying them all together)
`extrahosts` changed to `extra-hosts` (though note that the CLI equivalent is to have multiple `--add-hosts` parameters rather than supplying them all together)

docker/status_handler.go
nigeldeakin: This change was necessary to log the output that docker generates when it builds the image. These arrive in the `stream` element, not the `status` element of the JSON response. In this case the `status` element is not set.

docker/docker_client_fsouza.go
nigeldeakin: Possibly not: this is a copy of the copyright notice for the file from where this code was moved. But the code has been refactored into two separate files, which I would argue constitutes a change for the purposes of setting the copyright year.

nigeldeakin: But you can have the final say on this, and so I will remove the `2018`.

hatchan: I wasn't referring to the date. Mostly saying that I haven't reviewed the code.

nigeldeakin: Ha! The highlighted code was the copyright statement :-)

But no, I've just moved the `DockerClient` struct (which wraps the fsouza client struct), a constructor, and a few related utilities our of `docker.go` to a separate file `docker_client_fsouza.go`.

In the same way, I've created a new file `docker_client.go` which contains a constructor for the official docker client struct.

Eventually `docker_client_fsouza.go` will be redundant and can be removed.

PR #374, opened 2018-04-05
Fixes #353 - Updates to enable go build and wercker build work with golang 1.10

### LINKED ISSUE #353: go build failed with go1.10
go build passed with go1.9.3.darwin-amd64.pkg
go build with go1.10.darwin-amd64.pkg failed with following error:
# github.com/wercker/wercker/vendor/github.com/google/certificate-transparency/go/x509
vendor/github.com/google/certificate-transparency/go/x509/root_darwin.go:73: cannot use nil as type _Ctype_CFDataRef in assignment

lake-of-dreams:
Fixes [353](https://github.com/wercker/wercker/issues/353)
To make "go build" and "wercker build" work with golang 1.10 following changes were required to be done

- Updates to few cfssl, certificate-transparaency-go and related dependencies in vendor.json as per https://github.com/google/certificate-transparency-go/issues/153
- Updates to vendor.json to fetch golang.org/x/crypto/cryptobyte as mentioned in this post https://github.com/google/certificate-transparency-go/issues/153#issuecomment-370893611
- From golang version 1.10 - go vet also looks for package dependencies being available (as mentioned in release notes - https://golang.org/doc/go1.10#vet. Therefore an update was made to wercker.yml to invoke "govendor sync" before "govendor vet"
- In new go vet , Unkeyed composite literals are being flagged as mentioned in https://golang.org/cmd/vet/#hdr-Unkeyed_composite_literals - we use this at several places as mentioned below:

1. github.com/wercker/wercker/core
core/pipeline.go:194: github.com/wercker/wercker/util.Formatter composite literal uses unkeyed fields
2. github.com/wercker/wercker/docker
docker/options.go:49: github.com/wercker/wercker/util.Formatter composite literal uses unkeyed fields
docker/watchstep.go:267: github.com/wercker/wercker/util.Formatter composite literal uses unkeyed fields
3. github.com/wercker/wercker/cmd
cmd/main.go:932: github.com/wercker/wercker/util.Formatter composite literal uses unkeyed fields
cmd/runner.go:132: github.com/wercker/wercker/util.Formatter composite literal uses unkeyed fields
cmd/runner.go:468: github.com/wercker/wercker/util.Formatter composite literal uses unkeyed fields
 
Since it is a valid construct - to disable the warning added -compoistes=false to govendor vet. More [here](https://stackoverflow.com/questions/36273920/disable-go-vet-checks-for-composite-literal-uses-unkeyed-fields) 

-  Replaced libsystemd-dev for libsystemd-journal-dev due to following message after updating the box to golang 1.10
"Package libsystemd-journal-dev is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source
However the following packages replace it:
  libsystemd-dev"

Have been tested by running couple of builds against canary staged version of wercker built after these changes

### DIFF
--- a/wercker.yml
+++ b/wercker.yml
@@ -17,7 +17,7 @@ build:
         code: govendor sync
 
     - script:
-        name: go vet
+        name: go vet -composites=false
         code: govendor vet +local
 
     # - golint:
--- a/wercker.yml
+++ b/wercker.yml
@@ -18,7 +18,7 @@ build:
 
     - script:
         name: go vet -composites=false
-        code: govendor vet +local
+        code: govendor vet -composites=false +local
 
     # - golint:
     #     exclude: vendor
--- a/wercker.yml
+++ b/wercker.yml
@@ -14,11 +14,11 @@ build:
     
     - script:
         name: install dependencies
-        code: govendor sync 
+        code: govendor sync
 
     - script:
         name: go vet
-        code: govendor vet +local    
+        code: govendor vet +local
 
     # - golint:
     #     exclude: vendor
--- a/wercker.yml
+++ b/wercker.yml
@@ -11,14 +11,14 @@ build:
     - script:
         name: install govendor
         code: go get -u github.com/kardianos/govendor
+    
+    - script:
+        name: install dependencies
+        code: govendor sync 
 
     - script:
         name: go vet
         code: govendor vet +local    
-     
-    - script:
-        name: install dependencies
-        code: govendor sync
 
     # - golint:
     #     exclude: vendor
--- a/wercker.yml
+++ b/wercker.yml
@@ -1,4 +1,4 @@
-box: golang:1.7
+box: golang:1.10
 no-response-timeout: 15
 
 build:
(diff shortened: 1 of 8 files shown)


hatchan:
The new go vet issues, is that the difference between the following:

```
v := SomeStruct{var1, var2}
// vs
v:= SomeStruct{ prop1: var1, prop2: var2}
```

If so, then I would prefer that we do enable the check for this and fix the issue.

wercker.yml
```diff
@@ -1,28 +1,28 @@
...
     - install-packages:
         checkpoint: pkgs
-        packages: openssh-client pkg-config libsystemd-journal-dev
+        packages: openssh-client pkg-config libsystemd-dev
 
     - script:
         name: install govendor
         code: go get -u github.com/kardianos/govendor
+     
+    - script:
```
hatchan: Why is this moved above go vet? If there is no reason, then we should move it back. When we have a `go vet` issue, then we don't have to wait for `govendor sync` to finish.

wercker.yml
```diff
@@ -1,28 +1,28 @@
...
+        name: install dependencies
+        code: govendor sync
 
     - script:
-        name: go vet
-        code: govendor vet +local
+        name: go vet -composites=false
```
hatchan: See general comment.

wercker.yml
lake-of-dreams: see https://app.wercker.com/wercker/wercker/runs/build/5ac74a503ada9c00016bb5dc?step=5ac74a6b96f3f300015a7ad0 - as mentioned in PR description from golang 1.10 onwards go vet also looks for package dependencies and if that is not present throw errors - thats why I had moved govendor sync above it

wercker.yml
lake-of-dreams: Fixed all the places where it was giving error and removed -composites=false

PR #375, opened 2018-04-09
Fix #1740 - Correctly infer registry and repository from step input

lake-of-dreams:
Fixes [1740](https://github.com/wercker/backlog/issues/1740) 
Existing logic to infer the correct registry URL from repository and registry fields from step configuration was not consistent as mentioned in issue description. With this change

1. If no repository is specified, it is assumed that the user wants to push an image of current application for which  the build is running to wcr.io repository and therefore registry is inferred as
https://test.wcr.io/v2 and repository as test.wcr.io/<application-owner>/<application-name>
2. In case a repository is provided but no registry - registry is derived from the name of the domain (if any) from the registry - e.g. for a repository quay.io/<repo-owner>/<repo-name> - quay.io will be the registry host and https://quay.io/v2/ will be the registry url. In case the repository name does not contain a domain name - docker hub is assumed to be the registry and therefore any authorization with supplied username/password is carried out with docker hub.
3. In case both repository and registry are provided -
   3(a) - In case registry provided points to a wrong url - we use registry inferred from the domain name(if any) prefixed to the repository. However in this case if no domain name is specified in repository - we return an error since user probably wanted to use this repository with a different registry and not docker hub and should be alerted that the registry url is invalid.In case registry url is valid - we evaluate scenarios 4(b) and 4(c)
   3(b) - In case no domain name is prefixed to the repository - we assume repository belongs to the registry specified and prefix domain name extracted from registry.
   3(c) - In case repository also contains a domain name - we check if domain name of registry and repository are same, we assume that user wanted to use the registry host as specified in repository and change the registry to point to domain name present in repository. If domain names in both registry and repository are same - no changes are made.

- Changed method name to InferRegistryAndRepository and fixed the return types to correctly reflect what this method does
- Updated unit test cases for additional scenarios
- Fixed [comment on PR for 1741](https://github.com/wercker/wercker/pull/371#discussion_r179683496)
-  Added code to print repo:tag of pushed container in step logs
- vendor.json currently refers to [issue-1740 branch of docker-check-access](https://github.com/wercker/docker-check-access/tree/issue-1740)  which contain related changes as mentioned in this [PR](https://github.com/wercker/docker-check-access/pull/6) . Once that is merged - will change it back to refer to master of docker-check-access
- Have been tested using stage canary testing

### DIFF
--- a/docker/docker.go
+++ b/docker/docker.go
@@ -1153,11 +1153,12 @@ func (s *DockerPushStep) tagAndPush(imageID string, e *core.NormalizedEmitter, c
 				if statusMessage.Aux != nil && statusMessage.Aux.Tag == tag {
 					s.logger.Println("Pushed container:", s.repository, tag, ",Digest:", statusMessage.Aux.Digest)
 					e.Emit(core.Logs, &core.LogsArgs{
-						Logs: fmt.Sprintf("Pushed %s:%s", s.repository, tag),
+						Logs: fmt.Sprintf("\n Pushed %s:%s", s.repository, tag),
 					})
 					isContainerPushed = true
 				}
 
+				s
 			}
 			if !isContainerPushed {
 				s.logger.Errorln("Failed to push tag:", tag, "Please check log messages")
--- a/docker/docker.go
+++ b/docker/docker.go
@@ -936,50 +936,49 @@ func InferRegistryAndRepository(repository string, registry string, pipelineOpti
 		inferredRegistry = pipelineOptions.WerckerContainerRegistry.String()
 		_logger.Infoln("No repository specified - using " + inferredRepository)
 		_logger.Infoln("username/password fields are ignored while using wcr.io registry, supplied authToken (if provided) will be used for authorization to wcr.io registry")
-	} else {
-		// Docker repositories must be lowercase
-		inferredRepository = strings.ToLower(repository)
-		inferredRegistry = registry
-		x, _ := reference.ParseNormalizedNamed(inferredRepository)
-		domainFromRepository := reference.Domain(x)
-		registryInferredFromRepository := ""
-		if domainFromRepository != "docker.io" {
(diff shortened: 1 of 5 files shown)


docker/docker.go
```diff
@@ -904,25 +909,77 @@ func (s *DockerPushStep) buildAutherOpts(env *util.Environment) dockerauth.Check
...
 	if repository == "" {
-		repository = pipelineOptions.WerckerContainerRegistry.Host + "/" + pipelineOptions.ApplicationOwnerName + "/" + pipelineOptions.ApplicationName
-	}
-	// Docker repositories must be lowercase
-	repository = strings.ToLower(repository)
+		inferredRepository = pipelineOptions.WerckerContainerRegistry.Host + "/" + pipelineOptions.ApplicationOwnerName + "/" + pipelineOptions.ApplicationName
+		inferredRegistry = pipelineOptions.WerckerContainerRegistry.String()
+		_logger.Infoln("No repository specified - using " + inferredRepository)
+		_logger.Infoln("username/password fields are ignored while using wcr.io registry, supplied authToken (if provided) will be used for authorization to wcr.io registry")
+	} else {
```
hatchan: Close the previous if statement with `return inferredRepository, inferredRegistry, nil` and remove the `else`.

docker/docker.go
```diff
@@ -904,25 +909,77 @@ func (s *DockerPushStep) buildAutherOpts(env *util.Environment) dockerauth.Check
...
+				}
+
+			}
+		} else {
+			inferredRegistry = registryInferredFromRepository
 		}
+
 	}
-	return repository, opts
+	return
```
hatchan: Be explicit about the returned variables.

lake-of-dreams: fixed

docker/docker.go
lake-of-dreams: fixed

PR #378, opened 2018-04-10
Fix integration tests

nigeldeakin:
This PR fixes two unrelated failures in the integration tests.

The change to `tests/projects/docker-push/test.sh` is needed following the merge of PR https://github.com/wercker/wercker/pull/377 with the merge of PR https://github.com/wercker/wercker/pull/375 . Previously the registry property was not prepended to the repository property when tagging the image. Now it is.

The change to `tests/projects/multidb/wercker.yml` is needed because this test previously pulled `sutoiku/redis`, which no longer exists on Docker Hub.

### DIFF
--- a/tests/projects/docker-push/test.sh
+++ b/tests/projects/docker-push/test.sh
@@ -12,7 +12,7 @@ testDockerPush () {
   testDir=$testsDir/docker-push
   printf "testing %s... " "$testName"
   # this test will create an image with the following repository: should match the repository setting in wercker.yml
-  repo=hub.docker.com/test-docker-push
+  repo=my.registry.com/test-docker-push
   # stop any existing container started by a previous run
   docker kill ${testName}-container > /dev/null 2>&1
   # delete any existing image built by a previous run
--- a/tests/projects/docker-push/test.sh
+++ b/tests/projects/docker-push/test.sh
@@ -12,7 +12,7 @@ testDockerPush () {
   testDir=$testsDir/docker-push
   printf "testing %s... " "$testName"
   # this test will create an image with the following repository: should match the repository setting in wercker.yml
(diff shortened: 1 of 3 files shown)


tests/projects/docker-push/test.sh
```diff
@@ -12,7 +12,7 @@ testDockerPush () {
   testDir=$testsDir/docker-push
   printf "testing %s... " "$testName"
   # this test will create an image with the following repository: should match the repository setting in wercker.yml
-  repo=test-docker-push
+  repo=hub.docker.com/test-docker-push
```
hatchan: is this the docker image name the docker daemon locally has? That seems pretty odd.

lake-of-dreams: repository either should follow format:
 - <registry-domain-name>/<owner-name>/<repo-name>
for docker hub, registry name is actually registry.hub.docker.com and for docker hub repos, there is no need to specify <registry-domain-name> at all

nigeldeakin:
Here is the `wercker.yml`:
```
        registry: https://hub.docker.com
        repository: test-docker-push
        tag: latest   
```

Before https://github.com/wercker/wercker/pull/375, `docker images` displayed:
```
REPOSITORY                        TAG                 IMAGE ID    
test-docker-push   latest              3e8d9e43f91f    
``` 
now it displays
```
REPOSITORY                        TAG                 IMAGE ID    
hub.docker.com/test-docker-push   latest              3e8d9e43f91f    
```
This test uses `--docker-local` so doesn't actually push the image. It's just testing that the image is created (i.e. the docker commit) and that the labels and env vars are correct. Perhaps I should set `registry` to `https://my.registry.com`). The main thing is that the registry is prepended to the repository, when it wasn't before, and so this test needs to be changed to make it work.

PR #381, opened 2018-04-12
Fix 364 - Update azure client to allow push in all regions of azure

lake-of-dreams:
Update vendor.json with latest azure client.
changes to docker-check-access are in PR [7](https://github.com/wercker/docker-check-access/pull/7)
An unrelated change to add a newline after docker push message


lake-of-dreams:
Have been tested using canary

PR #382, opened 2018-04-12
Docker network

agnihotrigaurav:
Wercker cli use docker links for communication. It has been changed to use custom bridge network.

Proposal and design is described in https://confluence.oraclecorp.com/confluence/display/WRKR/Docker+Networks

### DIFF
--- a/cmd/flags.go
+++ b/cmd/flags.go
@@ -1,4 +1,4 @@
-//   Copyright 2016 Wercker Holding BV
+//   Copyright © 2016, 2018, Oracle and/or its affiliates.  All rights reserved.
 //
 //   Licensed under the Apache License, Version 2.0 (the "License");
 //   you may not use this file except in compliance with the License.
--- a/cmd/flags.go
+++ b/cmd/flags.go
@@ -50,7 +50,7 @@ var (
 		cli.IntFlag{Name: "docker-kernel-memory", Usage: "Set docker kernel memory limit in MB NOTIMPLEMENTED", Hidden: true},
 		cli.BoolFlag{Name: "docker-cleanup-image", Usage: "Remove image from the Docker when finished pushing them", Hidden: true},
 		cli.StringFlag{Name: "docker-network", Value: "", Usage: "Docker network name.", EnvVar: "DOCKER_NETWORK"},
-		cli.BoolFlag{Name: "docker-network-rm", Usage: "Remove passed docker network", Hidden: true),
+		cli.BoolFlag{Name: "docker-network-rm", Usage: "Remove passed docker network", Hidden: true},
 	}
 
 	// These flags control where we store local files
--- a/cmd/flags.go
+++ b/cmd/flags.go
@@ -49,7 +49,7 @@ var (
 		cli.IntFlag{Name: "docker-memory-reservation", Usage: "Set docker user memory soft limit in MB NOTIMPLEMENTED", Hidden: true},
 		cli.IntFlag{Name: "docker-kernel-memory", Usage: "Set docker kernel memory limit in MB NOTIMPLEMENTED", Hidden: true},
 		cli.BoolFlag{Name: "docker-cleanup-image", Usage: "Remove image from the Docker when finished pushing them", Hidden: true},
-		cli.StringFlag{Name: "docker-network", Value: "", Usage: "Docker network name.", EnvVar: "DOCKER_NETWORK"},
+		cli.StringFlag{Name: "docker-network", Value: "", Usage: "Docker network name.", Hidden: true},
 	}
 
 	// These flags control where we store local files
--- a/cmd/flags.go
+++ b/cmd/flags.go
@@ -50,7 +50,6 @@ var (
 		cli.IntFlag{Name: "docker-kernel-memory", Usage: "Set docker kernel memory limit in MB NOTIMPLEMENTED", Hidden: true},
 		cli.BoolFlag{Name: "docker-cleanup-image", Usage: "Remove image from the Docker when finished pushing them", Hidden: true},
 		cli.StringFlag{Name: "docker-network", Value: "", Usage: "Docker network name.", EnvVar: "DOCKER_NETWORK"},
-		cli.BoolFlag{Name: "docker-network-rm", Usage: "Remove passed docker network", Hidden: true},
 	}
 
 	// These flags control where we store local files
--- a/cmd/flags.go
+++ b/cmd/flags.go
@@ -49,6 +49,8 @@ var (
 		cli.IntFlag{Name: "docker-memory-reservation", Usage: "Set docker user memory soft limit in MB NOTIMPLEMENTED", Hidden: true},
 		cli.IntFlag{Name: "docker-kernel-memory", Usage: "Set docker kernel memory limit in MB NOTIMPLEMENTED", Hidden: true},
 		cli.BoolFlag{Name: "docker-cleanup-image", Usage: "Remove image from the Docker when finished pushing them", Hidden: true},
+		cli.StringFlag{Name: "docker-network", Value: "", Usage: "Docker network name.", EnvVar: "DOCKER_NETWORK"},
+		cli.BoolFlag{Name: "docker-network-rm", Usage: "Remove passed docker network", Hidden: true),
 	}
 
 	// These flags control where we store local files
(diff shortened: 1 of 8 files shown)


agnihotrigaurav:
Fixed compilation for issue compile linux_386

docker/box.go
```diff
@@ -119,24 +121,24 @@ func NewDockerBox(boxConfig *core.BoxConfig, options *core.PipelineOptions, dock
 	}, nil
 }
 
-func (b *DockerBox) links() []string {
-	serviceLinks := []string{}
+// func (b *DockerBox) links() []string {
```
hatchan: Remove this code

docker/box.go
```diff
@@ -344,17 +344,28 @@ func (b *DockerBox) getContainerName() string {
...
+			b.logger.Error("Error while creating network", err)
+			return nil, err
+		}
+	}
+
 	err := b.RunServices(ctx, env)
 	if err != nil {
 		return nil, err
 	}
+	dockerEnvVar, err := b.prepareSvcDockerEnvVar(env)
```
hatchan: err is not handled

docker/box.go
```diff
@@ -655,3 +674,66 @@ func (b *DockerBox) ExportImage(options *ExportImageOptions) error {
...
+			lowestPort := math.MaxInt32
+			var protLowestPort string
+			for k, _ := range container.Config.ExposedPorts {
+				s := strings.Split(string(k), "/")
+				x, _ := strconv.Atoi(s[0])
+				if lowestPort > x {
+					lowestPort = x
+					protLowestPort = s[1]
+				}
+				serviceEnv = append(serviceEnv, fmt.Sprintf("%s_PORT_%s_%s=%s://%s:%s", strings.ToUpper(serviceName), s[0], strings.ToUpper(s[1]), s[1], serviceIPAddress, s[0]))
```
hatchan: The first three parameters are always the same, so maybe store that in a string first: 
```
prefix:= fmt.Sprintf("%s_PORT_%s_%s", strings.ToUpper(serviceName), s[0], strings.ToUpper(s[1]))
```

Also might be useful to store s[0] and s[1] as variables, as I have no idea what currently is in there.

docker/box.go
```diff
@@ -389,9 +400,9 @@ func (b *DockerBox) Run(ctx context.Context, env *util.Environment) (*docker.Con
 
 	hostConfig := &docker.HostConfig{
 		Binds:        binds,
-		Links:        b.links(),
 		PortBindings: portBindings(portsToBind),
 		DNS:          b.dockerOptions.DNS,
+		NetworkMode:  dockerNetworkName,
```
hatchan: This seems a bit odd. What is NetworkMode and why is it being set to the network name?

docker/box.go
agnihotrigaurav: "ExposedPorts": {
       "5432/tcp": {}
  },
is the exposed ports. s[0] would have 5432 and s[1] would have tcp. 
changed s[0] and s[1] to variable name.

docker/box.go
```diff
@@ -344,17 +325,31 @@ func (b *DockerBox) getContainerName() string {
 
 // Run creates the container and runs it.
 func (b *DockerBox) Run(ctx context.Context, env *util.Environment) (*docker.Container, error) {
+	dockerNetworkName, custom := b.GetDockerNetworkName()
+	if custom == false {
+		_, err := b.createDockerNetwork(dockerNetworkName)
```
lake-of-dreams: I don't fully understand why do we need to take docker network name from user - the name can easily conflict per daemon and would result in many failures because different users can give same name and then per daemon it will fail when multiple pipelines are running on same daemon

docker/box.go
lake-of-dreams: This should not be necessary - you should be able to create a network and a container separately and then attach that container to the network separately - see docker network connect command

docker/box.go
```diff
@@ -655,3 +658,71 @@ func (b *DockerBox) ExportImage(options *ExportImageOptions) error {
...
+func (b *DockerBox) createDockerNetwork(dockerNetworkName string) (*docker.Network, error) {
+	b.logger.Debugln("Creating docker network")
+	client := b.client
+	return client.CreateNetwork(docker.CreateNetworkOptions{
+		Name:           dockerNetworkName,
+		CheckDuplicate: true,
+	})
+}
+
+// Prepares and return DockerEnvironment variables list created in case of DBlinks corresponding to each service.
```
lake-of-dreams: DBLink??

docker/box.go
```diff
@@ -655,3 +658,71 @@ func (b *DockerBox) ExportImage(options *ExportImageOptions) error {
...
+			var serviceIPAddress string
+			for _, v := range ns.Networks {
+				serviceIPAddress = v.IPAddress
+				break
+			}
+			serviceEnv = append(serviceEnv, fmt.Sprintf("%s_NAME=/%s/%s", strings.ToUpper(serviceName), b.getContainerName(), serviceName))
+			lowestPort := math.MaxInt32
+			var protLowestPort string
+			for k, _ := range container.Config.ExposedPorts {
+				exposedPort := strings.Split(string(k), "/") //exposedPort[0]=portNum and exposedPort[1]=protocal(tcp/udp)
```
lake-of-dreams: Not sure what this logic is trying to do - maybe add the algo to documentation of the method?

docker/box.go
```diff
@@ -486,6 +481,14 @@ func (b *DockerBox) Clean() error {
 		}
 	}
 
+	dockerNetworkName, custom := b.GetDockerNetworkName()
+	if custom == false {
+		err := client.RemoveNetwork(dockerNetworkName)
```
lake-of-dreams: Before removing a docker network - if there is any active container on that network - it will return an error - you should first inspect that network - remove all containers - and then remove network

docker/box.go
agnihotrigaurav: Its just for local testing.

docker/service.go
```diff
@@ -210,11 +212,20 @@ func (b *InternalServiceBox) Run(ctx context.Context, env *util.Environment, lin
...
+	}
+	endpointConfigMap := make(map[string]*docker.EndpointConfig)
+	endpointConfigMap[networkName] = endpointConfig
+
 	container, err := client.CreateContainer(
 		docker.CreateContainerOptions{
 			Name:       b.getContainerName(),
 			Config:     conf,
 			HostConfig: hostConfig,
+			NetworkingConfig: &docker.NetworkingConfig{
```
lake-of-dreams: In my opinion - we should not again couple docker network with box/service creation - we should create network separately and attach all the services followed by box to it ( for shortName thing you can use docker network connect --alias) - reason being if network creation fails for some reason - we should be able to fall back to bridge network and not fail the pipeline as such

cmd/flags.go
```diff
@@ -49,6 +49,7 @@ var (
 		cli.IntFlag{Name: "docker-memory-reservation", Usage: "Set docker user memory soft limit in MB NOTIMPLEMENTED", Hidden: true},
 		cli.IntFlag{Name: "docker-kernel-memory", Usage: "Set docker kernel memory limit in MB NOTIMPLEMENTED", Hidden: true},
 		cli.BoolFlag{Name: "docker-cleanup-image", Usage: "Remove image from the Docker when finished pushing them", Hidden: true},
+		cli.StringFlag{Name: "docker-network", Value: "", Usage: "Docker network name.", Hidden: true},
```
lake-of-dreams: even for a cli user - I think the logic should create a network and spit its name in output - if at all user wants to inspect the network during pipeline build - will help maintain uniqueness at least on our end

docker/box.go
agnihotrigaurav: Separate call to connect is not required in network is already created and its name is passed as NetworkMode

docker/box.go
agnihotrigaurav: Done.

docker/service.go
agnihotrigaurav: every build should run on separate network.

docker/box.go
agnihotrigaurav: Updated comments for same.

docker/box.go
agnihotrigaurav: Updated comments.

PR #384, opened 2018-04-12
Bihaber external runner

bhabermaas:
This update contains changes to the Wercker CLI to support external runners.  An external runner container (containing a modified kiddie-pool) is started by the runner start command.  Once the container is started, it runs independently to accept jobs from dispatch.  Dispatch will filter jobs available to the runner according to runner start parameters. 

1. Add runner command and subcommands 'start', 'stop' 'status'. 
2. Add External runner controller to handle the runner subcommands.


external/controller.go
```diff
@@ -0,0 +1,332 @@
...
+	args := []string{}
+	labels := []string{}
+	volumes := []string{}
+
+	labels = append(labels, fmt.Sprintf("runner=/wercker-external-runner-%s", cp.Basename))
+	if cp.GroupName != "" {
+		labels = append(labels, fmt.Sprintf("runnergroup=%s", cp.GroupName))
+	}
+
+	volumes = append(volumes, "/Users/bihaber/wercher:/var/lib/wercker:rw")
```
gmcadoo: Not sure we want to have your hardcoded path for the volumes in here.

You also spelled wercker wrong (not sure if that was on purpose or not.

cmd/flags.go
```diff
@@ -250,6 +250,25 @@ var (
...
+	ExternalRunnerCommonFlags = []cli.Flag{
+		cli.StringFlag{Name: "name", Usage: "specify a unique name for active runner(s))"},
+		cli.StringFlag{Name: "group", Usage: "specify the group name for starting runner(s)"},
+	}
+
+	ExternalRunnerStartFlags = []cli.Flag{
+		cli.StringFlag{Name: "name", Usage: "specify an instance name for runner(s)"},
+		cli.StringFlag{Name: "group", Usage: "specify the group name for runner(s)"},
+		cli.StringFlag{Name: "orgs", Usage: "comma separated list of organization names"},
+		cli.StringFlag{Name: "apps", Usage: "comma separated list of application names"},
```
mgianatagh: The format of the app names is org-name/app-name

cmd/flags.go
```diff
@@ -250,6 +250,25 @@ var (
...
+		cli.StringFlag{Name: "workflows", Usage: "comma separated list of workflows"},
```
mgianatagh: The format of the workflow name is org-name/app-name/workflow-name

external/controller.go
```diff
@@ -0,0 +1,332 @@
...
+		cmd, err := cp.createTheRunnerCommand(runnerName)
+		if err == nil {
+			cp.startTheContainer(runnerName, cmd)
+			ct++
+		}
+	}
+}
+
+// Create the command to run the external runner in a container.
+// Temporarily just read in the arguments from the runner.args file
```
mgianatagh: Stale comment?

external/controller.go
```diff
@@ -0,0 +1,332 @@
...
+	log.Print(message)
+	return nil
+}
+
+// Execute the docker command
+func runDocker(args []string) error {
+	dockerCmd := exec.Command("docker", args...)
+	// run using a pseudo-terminal so that we get the nice docker output :)
+	err := dockerCmd.Start()
+	//_, err := pty.Start(dockerCmd)
```
mgianatagh: Can the commented out code in this func be removed?

external/controller.go
```diff
@@ -0,0 +1,328 @@
...
+	"log"
+	"os"
+	"os/exec"
+	"strings"
+	"time"
+
+	"github.com/fsouza/go-dockerclient"
+)
+
+// RunnerParams - The parameters that drive the control of Docker
```
hatchan: Write documentation as normal sentences. Ie. "RunnerParams are the parameters..."

external/controller.go
```diff
@@ -0,0 +1,328 @@
...
+// containers twhere the external runner executes. This structure is
```
hatchan: twhere -> where

external/controller.go
```diff
@@ -0,0 +1,328 @@
...
+	return &RunnerParams{
+		ImageName: "oracle/wercker/runner:latest",
+	}
+}
+
+// RunDockerController - This is commander-in-chief of external runners. It is called from
+// Wercker CLI to start or stop external runners. The Wercker CLI builds the RunnParams and
+// calls this function.
+func (cp *RunnerParams) RunDockerController(statusOnly bool) {
+
```
hatchan: Remove these empty lines after func deceleration (applies to all files)

external/controller.go
```diff
@@ -0,0 +1,328 @@
...
+	// When no instance name was supplied, use the hostname
+	cp.Basename = cp.InstanceName
+	if cp.InstanceName == "" {
+		hostName, err := os.Hostname()
+
+		if err != nil {
+			log.Print(fmt.Sprintf("unable to access hostname: %s", err))
```
hatchan: We should resuse the logging framework that the rest of wercker cli is using.
Then we should use log levels (this sounds like a error).
And we should use `Errorf`, instead of calling `fmt.Sprintf`
(applies to other locations as well)

external/controller.go
```diff
@@ -0,0 +1,328 @@
...
+			return
+		}
+		cp.Basename = hostName
+	}
+
+	endpoint := "unix:///var/run/docker.sock"
```
hatchan: Wercker already has the docker client as a flag, we should not hardcode it here.

external/controller.go
```diff
@@ -0,0 +1,328 @@
...
+	PollFreq     int    // Polling frequency
+	// following values are set during processing
+	Basename string // base name for container creation
+	client   *docker.Client
+}
+
+// NewDockerController -
+func NewDockerController() *RunnerParams {
+	return &RunnerParams{
+		ImageName: "oracle/wercker/runner:latest",
```
hatchan: This doesn't seem like a valid Docker Image

external/controller.go
```diff
@@ -0,0 +1,328 @@
...
+	}
+
+	// runners contains the containers running for this external runner
+	if cp.ShutdownFlag == true {
+		// Go handle shutdown of our runners.
+		cp.shutdownRunners(runners)
+		return
+	}
+
+	if statusOnly == true {
```
hatchan: Remove `== true`

external/controller.go
```diff
@@ -0,0 +1,328 @@
...
+}
+
+// Starting runner(s).  Initiate a container to run the external runner for as many times as
+// specified by the user.
+func (cp *RunnerParams) startTheRunners() {
+
+	if cp.BearerToken == "" {
+		// Check if token is supplied in the environment and pick it up from
+		// there.
+		token := os.Getenv("WERCKER-RUNNER-TOKEN")
```
hatchan: This is different from what is defined in the flag, which is "WERCKER_RUNNER_TOKEN". Whereas this is using dashes. 

if the only correct way is `WERCKER_RUNNER_TOKEN` then we do not have to do the code that is defined here, as the flag library will already use the `WERCKER_RUNNER_TOKEN` if no flag is defined.

hatchan:
Also for future reference, please us "Reviewer" instead "Assigned"

external/controller.go
bhabermaas: Please clarify what should be a valid Docker image for external runner.  I don't know if there a standard but I picked a name that is similar to other oracle docker image names.

hatchan: Well this looks like it would resolve to the official docker hub registry since there is no domain in there, but then it would fail since there is a / in the repo name, which I think is not allowed.

bhabermaas: I'm waiting for a ruling on what the image should be called. For the time being this is working.

external/controller.go
bhabermaas: fixed

bhabermaas: Fixed

PR #385, opened 2018-04-13
Migrate to official docker client for golang

nigeldeakin:
This PR changes some, but not all, of the wercker CLI code to use the official docker client instead of the fsouza docker client. See issue https://github.com/wercker/backlog/issues/1698


nigeldeakin:
This is work in progress.

nigeldeakin:
The most significant change in the preceding commits is to `docker/artifact.go`. This is the code that downloads artifacts from the pipeline container. This is used in several places: at the end of a pipeline, before a `docker-build` step, and (if the `--artifacts` flag is specified) at the end of every step. This code now uses the official docker client. 

Fortunately there are some unit tests for this code, which I have updated to test the new implementation. Note that since these particular unit tests require a docker daemon to be available they are automatically skipped when the unit tests are run as part of a wercker build, and you have to run them locally using `go test`.

I have also changed a number of minor API calls to use the official docker client. 

Most functions in the official docker client require a `Context` to be passed in. I have changed each command handler in `main.go` to create a `Context` (some commands already did this) and added additional parameters to intermediate functions to allow this to be passed down to the docker client.

PR #390, opened 2018-04-20
--store=oci

shihchang:
Add --store option to wercker-cli, supported values are "oci" and "s3" which --store=s3 is equivalent to existing --store-s3[=true]. The shouldStoreS3, used internally not expose to cli options, is renamed to shouldStore and --store-s3, --store=s3, or --store=oci will turn shouldStore=true.


cmd/flags.go
```diff
@@ -165,6 +168,16 @@ var (
 		cli.StringFlag{Name: "aws-region", Value: "us-east-1", Usage: "AWS region to use for artifact storage."},
 	}
 
+	//OCI ObjectStore options
+	OCIObjStoreFlags = []cli.Flag {
```
desagar: AWSFlags includes S3 access key and secret key, but OCI object store flags does not include the OCI access details (tenancy ocid, user ocid, fingerprint, private key and region). Did you omit them simply because there are too many? If so, I think in the docs for "store" flag, we should provide a full list of the env vars needed.

core/artifact.go
```diff
@@ -40,7 +43,12 @@ type Artifact struct {
 
 // URL returns the artifact's S3 url
```
desagar: comment should be updated to S3 or OCI obj store

core/ocistore.go
```diff
@@ -0,0 +1,92 @@
...
+	"os"
+	"context"
+)
+
+/*
+OciEnvVarPrefix is the prefix to use for all environment variables needed by the OCI SDK
+ */
+const OciEnvVarPrefix  = "wkr"
+
+// NewS3Store creates a new S3Store
```
desagar: update comment (and other method level comments)

core/ocistore.go
```diff
@@ -0,0 +1,92 @@
...
+	logger  *util.LogEntry
+	options *OciOptions
+}
+
+// StoreFromFile copies the file from args.Path to options.Bucket + args.Key.
+func (this *OciStore) StoreFromFile(args *StoreFromFileArgs) error {
+	if args.MaxTries == 0 {
+		args.MaxTries = 1
+	}
+	configProv := ocisdkcomm.ConfigurationProviderEnvironmentVariables(OciEnvVarPrefix, "");
```
desagar: I used the environment variable provider in Grappler, because Wercker was the only user. Since the wercker CLI can be used by end users, now I am thinking we should probably use the DefaultConfigProvider, I think we should use the DefaultConfigProvider, which searches the <user home>/.oci/config (which is the standard location for OCI config) and then another location, and finally looks for env vars. See here for the OCI SDK DefaultConfigProvider code: https://github.com/oracle/oci-go-sdk/blob/master/common/client.go We can then refer the user to the standard OCI docs to show them how to set the config.

cmd/flags.go
shihchang: Those comes from env vars, and the oci client reads those directly from env variables. I am not sure if it is correct to put them into the flags, but it would be redundant. If oci client support passing those as parameters then we may move those to flags/options.

shihchang: I changed the configProvider to use a composite config provider of ConfigurationProviderEnvironmentVariables and DefaultConfigurationProvider, so it is configProv {[Configuration provided by environment variables prefixed with: wkr {[Configuration provided by file: ~/.oci/config Configuration provided by file: ~/.oraclebmc/config Configuration provided by environment variables prefixed with: TF_VAR]}]}

core/ocistore.go
```diff
@@ -0,0 +1,96 @@
...
+	options *OciOptions
+}
+
+// StoreFromFile copies the file from args.Path to options.Bucket + args.Key.
+func (this *OciStore) StoreFromFile(args *StoreFromFileArgs) error {
+	if args.MaxTries == 0 {
+		args.MaxTries = 1
+	}
+	configProviders := []ocisdkcomm.ConfigurationProvider {
+		ocisdkcomm.ConfigurationProviderEnvironmentVariables(OciEnvVarPrefix, ""),
```
desagar: I understand you have this composing provider in order to support OciEnvVarPrefix, but I am wondering if we need to support that prefix since we just introduced it. The only other place which uses that is grappler which is not exposed to end user. Could we make do with just DefaultConfigProvider? What is your opinion?

cmd/flags.go
```diff
@@ -103,6 +103,9 @@ var (
 			(~/.aws/config, AWS_SECRET_ACCESS_KEY, etc), or from the --aws-secret-key and
 			--aws-access-key flags. It will upload to a bucket defined by --s3-bucket in
 			the region named by --aws-region`},
+		cli.StringFlag{Name: "store", Value: "",
```
desagar: We need more documentation here for the end user on env vars and config file. For the config file option, you can provide a link to the OCI docs on configuration file here: https://docs.us-phoenix-1.oraclecloud.com/Content/API/Concepts/sdkconfig.htm

hatchan: These flags should only be used (or rather supported) by the hosted wercker platform, so I think we should make these flags private/hidden and possibly not include too much details in here.

hatchan: Also can you check if there is a other property for the a longer description. Having Usage just be a single line would be more readable.

cmd/flags.go
```diff
@@ -171,6 +180,16 @@ var (
 		cli.StringFlag{Name: "keen-project-id", Value: "", Usage: "Keen project id.", Hidden: true},
 	}
 
+	//OCI ObjectStore options
+	OCIObjStoreFlags = []cli.Flag {
+		cli.StringFlag{
+			Name:  "namespace, n", Usage: "Specify the object store namespace",
```
hatchan: I would prefer not having these short alias (`n` and `b`) for flags which are most likely are not going to be used by the majority of the users.

core/ocistore.go
```diff
@@ -0,0 +1,96 @@
...
+
+import (
+	"github.com/wercker/wercker/util"
+	ocisdkcomm "github.com/oracle/oci-go-sdk/common"
+	ocisdkstorage "github.com/oracle/oci-go-sdk/objectstorage"
+	"os"
+	"context"
+)
+
+/*
```
hatchan: this can just be a `//` comment

cmd/flags.go
shihchang: I change all the oci related options to Hidden. I follow the --store-s3 way to have a longer or more informative description. But when I ran --help to show it, I found it actually shows only the 1st line and I cannot figure out how to make the --store-s3 show the longer/multi-lines description.

core/ocistore.go
shihchang: fixed.

PR #391, opened 2018-04-20
Replace hyphen with underscore for services in env

jasisk:
Mentioned this in your slack channel.

When docker injects the environment variables for linked containers, it replaces hyphens with underscores. This is because hyphens are invalid environment variable characters in most shells.

Wercker no longer appears to replicate this logic. There is a workaround to resolve the issue with a change to a project's `wercker.yml`:
```
services:
  - some-namespace/some-service
```

... replace with ...

```
services:
  - id: some-namespace/some-service
    name: some_service
```

This PR does the same by replacing hyphens with underscores in the service alias. It's not great but it at least demonstrates the fix.


markxnelson:
@hatchan please review this PR

PR #399, opened 2018-04-26
Convert docker-push to use the official docker client

nigeldeakin:
This change converts `docker-push` and `docker-scratch-push` to use the official docker client.

Here are a few specific which might help in the review:

* Most of the changes are in `docker.go`. Note that as this is a large file GitHub does not display the changes unless you click on the comparison pane.

* The `ImagePush` function returns output  in JSON format. I removed the custom code that parsed this output and replaced it with a call to the standard `EmitStatus` function, as used with other API calls. 

* I have extended the `EmitStatus` function to detect whether the JSON includes an `error` element. If it does the push is considered to have failed. 

* I have removed the code that tried to work out whether or not the push had failed by checking for a success message. I haven't found any test cases where this is necessary.

* I have run the test case for https://jira.oraclecorp.com/jira/browse/WRKR-54 (Failed docker-push generates no output in app.wercker.com) and confirms it reports the same error message as currently. So this change does not break that fix. (In this issue, if the `repository` property is set to `docker.io//docker-build-golang` the push step fails with the  error message  `docker.io//docker-build-golang is not a valid repository, error while validating repository name: invalid reference format`)

* I have run the test case for https://github.com/wercker/backlog/issues/1959 ( docker-push fails in prod with "Code: , Message: Index response didn't contain any endpoints"). and confirms it reports the same error message as currently. So this change does not break that fix. (In this issue, the `registry` property is set to a bad value, `https://hub.docker.com`. The step fails with `Error interacting with this repository: hub.docker.com/nigeldeakin/docker-build-golang Not authorized to push to https://hub.docker.com/v1/nigeldeakin/docker-build-golang. Please check username/password and registry/repository values`). 

* I also tested a bad password. This gives `Error interacting with this repository: docker.io/nigeldeakin/docker-build-golang Not Authorized to access the repository` as before.

* I have removed the `--force-tags` parameter as neither the old nor new docker clients support it, and it does not appear to be a feature of the underlying docker API either. This parameter is not mentioned in the wercker documentation.

* I have moved `RequireDockerEndpoint` from `docker.go` to `docker_client.go`. The intention is for  `docker_client.go` to contain generic code and for`docker.go` to only contain code for pushing. I plan to rename `docker.go` to `docker_pull.go` in a future PR.


docker/docker.go
```diff
@@ -508,6 +444,12 @@ func NewDockerPushStep(stepConfig *core.StepConfig, options *core.PipelineOption
 	}, nil
 }
 
+// Just to make life easier
```
lake-of-dreams: Why are we ignoring error here?

docker/artifact.go
```diff
@@ -97,13 +96,13 @@ func (a *Artificer) Upload(artifact *core.Artifact) error {
 
 // DockerFileCollector impl of FileCollector
 type DockerFileCollector struct {
-	client      *client.Client
+	client      *OfficialDockerClient
```
lake-of-dreams: After this change - will there be any other place still in code which will be using old client? If not - we should consider removing "Official" thing from the name imo

lake-of-dreams:
Title of PR should be Covert "docker-push" ...

docker/docker.go
```diff
@@ -846,45 +784,41 @@ func (s *DockerPushStep) Execute(ctx context.Context, sess *core.Session) (int,
...
 	// if image is specified then it is assumed to be the name or ID of an existing image
 	// if image is not specified then create a new image by committing the pipeline container
 	if imageID == "" {
-		commitOpts := docker.CommitContainerOptions{
-			Container:  containerID,
-			Repository: s.repository,
-			Author:     s.author,
-			Message:    s.message,
-			Run:        &config,
-			Tag:        s.tags[0],
```
lake-of-dreams: I think this existing code had bug which we are propagating - there can be multiple tags - here it seems to only take first tag ( s.tags[0]) , commit it and then call s.tagAndPush() - s.tagAndPush() pushes each tag present in s.tags but here in the code - it seems it only "commits" tags[0]. This should be changes to commit each tag from s.tags[] array and then call s.tagAndPush that will then push all of those tags

docker/docker.go
```diff
@@ -897,91 +831,58 @@ func (s *DockerPushStep) buildTags() []string {
...
-				s.logger.Errorln("Failed to push tag:", tag, "Please check log messages")
-				return 1, errors.New(NoPushConfirmationInStatus)
+			response, err := client.ImagePush(ctx, target, imagePushOptions)
+			if err != nil {
+				s.logger.Errorln("Failed to push:", err)
+				return 1, err
 			}
+			defer response.Close()
 
+			err = EmitStatus(e, response, s.options)
```
lake-of-dreams: So we are not checking status message anymore - I am assuming we have some status checking enabled in EmitStatus now which will catch any error at the time of push?

docker/artifact.go
nigeldeakin: There will still be lots of places where the fsouza client will continue to be used so we will need to distinguish between the two for some time (and it is better if the different is very obvious)

Once there are no more references to the old client I will remove it from the dependencies and rename all the places that use the prefix "official". I will also delete `docker_client_fsouza.go` completely.

docker/docker.go
nigeldeakin: Yes, I mentioned that (briefly) in the description: I have extended the `EmitStatus` function to detect whether the JSON includes an error element. If it does the push is considered to have failed and an error is returned.

nigeldeakin: Please see the changes to `docker/status_handler.go`

docker/docker.go
nigeldeakin: I've removed the function and moved its code inline. The error is no longer ignored.

docker/status_handler.go
```diff
@@ -41,9 +41,13 @@ type JSONMessageProcessor struct {
 
 // ProcessJSONMessage will take JSONMessage m and generate logs based on the
 // message and previous messages.
-func (s *JSONMessageProcessor) ProcessJSONMessage(m *jsonmessage.JSONMessage) string {
+func (s *JSONMessageProcessor) ProcessJSONMessage(m *jsonmessage.JSONMessage) (string, error) {
+	if m.Error != nil {
+		return m.Error.Error(), m.Error
```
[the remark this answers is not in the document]
nigeldeakin: Done

docker/docker_client.go
```diff
@@ -41,5 +55,26 @@ func NewOfficialDockerClient(options *Options) (*client.Client, error) {
 	if err != nil {
 		return nil, err
 	}
-	return dockerClient, nil
+	return &OfficialDockerClient{Client: dockerClient}, nil
+}
+
+func RequireDockerEndpoint(ctx context.Context, options *Options) error {
```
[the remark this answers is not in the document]
nigeldeakin: Added `// RequireDockerEndpoint attempts to connect to the specified docker daemon and returns an error if unsuccessful` . This isn't a new function.

PR #405, opened 2018-05-03
Docker files integration

agnihotrigaurav:
Changes are as follows.
01) Merged docker-networks, docker-run and docker kill.
02) Jira issue 161
03) Jira issue 162

Staging testing.
https://dev.wercker.com/gauagnih/docker-networks1/runs/build/5aeaba6f8b18e100015468fe?step=5aeaba6f8b18e10001546903


docker/docker_network.go
```diff
@@ -0,0 +1,191 @@
...
+
+// Generate docker network name and check if same is already in use. In case name is already in use then it regenerate it upto 3 times before throwing error.
+func (b *DockerBox) prepareDockerNetworkName() (string, error) {
+	generator := shortid.Generator()
+	client := b.client
+
+	for i := 0; i < 3; i++ {
+		dockerNetworkName := generator.Generate()
+		dockerNetworkName = "w-" + dockerNetworkName
+		network, _ := client.NetworkInfo(dockerNetworkName)
```
lake-of-dreams: Is it documented somewhere that in which situation it can generate same shortId - if yes we should reference it here

docker/docker_run.go
```diff
@@ -0,0 +1,336 @@
...
+	}
+
+	client, err := NewDockerClient(s.dockerOptions)
+	if err != nil {
+		return "", err
+	}
+
+	// local image should exists with a prepend of build id.
+	image, err := client.InspectImage(s.options.RunID + i)
+	if err != nil {
```
lake-of-dreams: why return i if there is an error?

test-all.sh
```diff
@@ -111,6 +113,47 @@ testScratchPush () {
   fi
 }
 
+testDockerNetworks () {
```
snanurag: shouldn't you write different files for docker-network and docker-kill tests? like everyone else has done?

nigeldeakin: I don't think we should be adding new tests to `test-all.sh`. Add the test script to the test directory and invoke it from `test-all.sh` (see the test for `docker-build`)

tests/projects/docker-n-networks/wercker.yml
```diff
@@ -0,0 +1,102 @@
+box: golang
```
nigeldeakin: Why is the directory called `docker-n-networks` rather than `docker-networks`. The `-n` (which in another test means "and") doesn't seem appropriate here.

agnihotrigaurav: Changed to docker-netowks

test-all.sh
agnihotrigaurav: Done.

docker/docker_network.go
agnihotrigaurav: "https://github.com/SKAhack/go-shortid" does not have explicit documentation. Similar library("https://github.com/teris-io/shortid" ) exist which says in rare circumstances its possible. If such circumstance exist then its better to generate another name rather than throwing error. Moreover need to log another bug which says to use runID instead of generation docker network name once customer moves to newer image.

nigeldeakin:
Thanks for the test changes. (Since I already submitted a review I don't see an option to approve it at this stage.)

docker/docker_run.go
```diff
@@ -0,0 +1,336 @@
...
+		return i, nil
+	} else {
```
hatchan: Remove this else, as the previous line is already returning

docker/docker_run.go
```diff
@@ -0,0 +1,336 @@
...
+}
+
+// The most important point in this function is that the required image is not pulled at first but the pipeline run-id is appended to
+// the input image name and its existence is checked locally. If that fails only then the input image is pulled from its respective registry.
+// This is done like this because docker-run is usually run in integration with docker-build. And, docker-build image is present in the
+// docker-deamon by <BuildId><image> name.
+func getCorrectImageName(env *util.Environment, s *DockerRunStep) (string, error) {
+	i := env.Interpolate(s.data["image"])
+	if i == "" {
+		err := fmt.Errorf("\"image\" is a required field. Please provide an image.")
```
hatchan: The last part is a bit redundant, so you can just remove everything after "required field".

docker/docker_run.go
```diff
@@ -0,0 +1,336 @@
...
+			s.EntryPoint = parts
+		} else {
+			return err
+		}
+	}
+
+	if envi, ok := s.data["env"]; ok {
+		parsedEnv, err := shlex.Split(envi)
+
+		if err == nil {
```
hatchan: Invert this if statement to match every other err handle statement in go (happens on multiple lines in this file, please change them all)

docker/docker_run.go
```diff
@@ -0,0 +1,336 @@
...
+}
+
+// NewDockerRunStep is a special step for doing docker runs. "image" is the required property for this step. It first checks if the input
+// "image" is build by docker-build and resided in docker-deaon. If not, then it uses the DockerBox logic to fetch the image and starts
+// a new container on that image. The new container is started on the Box network and cleaned up at the end of pipeline.
+func NewDockerRunStep(stepConfig *core.StepConfig, options *core.PipelineOptions, dockerOptions *Options) (*DockerRunStep, error) {
+	name := "docker-run"
+	displayName := "docker run"
+	originalContainerName := name
+	if stepConfig.Name != "" {
```
hatchan: invert this to be `if stepConfig.Name == "" {` and then remove the else statement.

docker/docker_run.go
agnihotrigaurav: Done

docker/docker_kill.go
```diff
@@ -0,0 +1,129 @@
...
+	removeContainerOpts := docker.RemoveContainerOptions{
+		ID: containerToKill,
+	}
+	s.logger.Debugln("Remove container:", containerToKill)
+	err = client.RemoveContainer(removeContainerOpts)
+	if err != nil {
+		s.logger.Errorln("Failed to remove container", err)
+		return -1, err
+	}
+	s.logger.WithField("container-name", s.containerName).Debug("Docker-kill completed")
```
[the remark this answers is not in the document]
agnihotrigaurav: Changed container-name to containerName

docker/docker_kill.go
```diff
@@ -0,0 +1,129 @@
...
+	// TODO(termie): could probably re-use the tansport's client
+	client, err := NewDockerClient(s.dockerOptions)
+	if err != nil {
+		return 1, err
+	}
+	containerToKill := s.options.RunID + s.containerName
+	killOpts := docker.KillContainerOptions{
+		ID: containerToKill,
+	}
+	s.logger.Debugln("kill container:", containerToKill)
```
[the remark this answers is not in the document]
agnihotrigaurav: Done.

docker/docker_run.go
agnihotrigaurav: Done.

docker/docker_run.go
agnihotrigaurav: Done.

docker/docker_run.go
agnihotrigaurav: Done.

PR #420, opened 2018-06-01
Docker push escape dollar char

agnihotrigaurav:
Solution
If any value contains $ sign and variable after $ sign does not exist in environment variables list, in that case $ sign would be escaped else previous behavior continues i.e. replace its value with variable.

Ex. 
case 1 :

DOCKER_IO_PASSWORD=PREFIX$SUFFIX
No variable exist with $SUFFIX in environment variable list.
-Then DOCKER_IO_PASSWORD=PREFIX\$SUFFIX
case 2:

DOCKER_IO_PASSWORD=PREFIX$SUFFIX
SUFFIX=something
-Then DOCKER_IO_PASSWORD=PREFIXsomething

-------------------------------------------------------------------------------------------------------------------
Staging test

https://dev.wercker.com/gauagnih/docker-push/runs/build/5b1114bb7b434d00013fb98d?step=5b1114da78eb58000117ebf9


util/environment.go
```diff
@@ -80,7 +80,21 @@ func (e *Environment) Get(key string) string {
 func (e *Environment) Export() []string {
 	s := []string{}
 	for _, key := range e.Order {
-		s = append(s, fmt.Sprintf(`export %s=%q`, key, e.Map[key]))
```
lake-of-dreams: There should be unit tests for all the scenarios

agnihotrigaurav: Added test case to code

util/environment.go
```diff
@@ -80,7 +80,21 @@ func (e *Environment) Get(key string) string {
...
+		value := e.Map[key]
+		if strings.Contains(value, "$") {
+			data := strings.Split(value, "$")
+			for i := 1; i < len(data); i++ {
+				if _, ok := e.Map[data[i]]; ok {
+					data[i] = "$" + data[i]
```
hatchan: I don't see a test for this scenario.

hatchan: Is the following notation working: `${test}` ?

agnihotrigaurav: Added a test case for this particular scenario. 
Updated code to make it work for ${test} notation.

markxnelson:
I pushed a new test, that I think highlights a problem with this approach:

```--- FAIL: TestEnvironmentSuite (0.00s)
    --- FAIL: TestEnvironmentSuite/TestExport (0.00s)
        Error Trace:    environment_test.go:118
    	Error:      	Not equal: 
    	            	expected: []string{"export TEST=\"test\"", "export DOCKER_IO_USER=\"user\"", "export DOCKER_IO_PASSWORD=NOT\\$ME"}
    	            	actual: []string{"export TEST=\"test\"", "export DOCKER_IO_USER=\"user\"", "export DOCKER_IO_PASSWORD=NOT\\\\$ME"}
    	            	
    	            	Diff:
    	            	--- Expected
    	            	+++ Actual
    	            	@@ -3,3 +3,3 @@
    	            	  (string) (len=28) "export DOCKER_IO_USER=\"user\"",
    	            	- (string) (len=33) "export DOCKER_IO_PASSWORD=NOT\\$ME"
    	            	+ (string) (len=34) "export DOCKER_IO_PASSWORD=NOT\\\\$ME"
    	            	 }
FAIL
FAIL	github.com/wercker/wercker/util	0.041s```

agnihotrigaurav:
Hello Mark,

Seems Expected result you have mentioned in test needs modification. ("DOCKER_IO_PASSWORD=NOT\\$ME")  backward slash would escape backward slash only. It has nothing to do with solution proposed. 
**Solution proposed.**
If there is $ sign and any thing after can not be evaluated then $ sign is escaped. 
**Previous behavior** 
If there is $ sign and anything after can not be evaluated then $ sign and any character after it would be truncated.


  
I performed below test  on staging which is working perfectly.(https://dev.wercker.com/gauagnih/docker-push/runs/build/5b2353c5c2c932000140e205)

docker repository password : ChangeMe\$AB

Environment variables 
1) TEST             : test
2) USERNAME  : gauravaagnihotri
3) PASSWORD : ChangeMe\\$AB      (UI would not allow single backslash)

docker push step
------------------------------------------------------------------------------------------------------------
    - internal/docker-push: 
        image-name: my-new-image
        username: $USERNAME # Docker Hub username. When using CLI, set using "export X_USERNAME=<username>"  
        password: $PASSWORD # Docker Hub password. When using CLI, set using "export X_PASSWORD=<password>" 
        repository: docker.io/$USERNAME/docker-build-golang
        tag: latest    
------------------------------------------------------------------------------------------------------------

agnihotrigaurav:
I doubt there will be anyone who will be using this functionality that in case there is a $ sign and characters after it can not be evaluated then $ sign and characters after it would be truncated. 

I agree there is a change in user experience corresponding to $ sign. We should educate user for the same.

If you still feels approach is not correct kindly suggest new approach.

markxnelson:
I reviewed this with @gilbode . We are not sure this is the best place to implement this, but we are ok to go ahead and merge this now, and we can iterate on it later if needed.

markxnelson:
Please make sure to merge and deploy this at the start of your day, just in case it does break anyone's build.  And better arrange to coordinate with someone in US to watch for issues after you go offline at the end of your day too.

ace03uec:
@agnihotrigaurav @snanurag : This is leading to behavior breaking for users.

gregoriokusowski:
Hi. I'm using the kubectl step, and I'm using `$` to refer to some folders.
Since I'm not getting the quotes around my command, 

The kubectl command gets exported as `export WERCKER_KUBECTL_COMMAND=apply -f \$ZZZ/YYY/` and fails with `/bin/sh: export: line 126: -f: bad variable name`.

Can we quote it everytime or revert this?

nigeldeakin:
We agree this change caused a regression so we've reverted the released version and will formally revert it in git shortly.

PR #421, opened 2018-06-04
Wrkr 389 - Changes for Accessing RDD API and injecting RDD url in pipeline builds

lake-of-dreams:
- Added flag 'rdd-dervice-uri' which will point to RDD API Service and optional 'rdd-provision-timeout' which represents maximum no. of seconds to poll the API service
- Support for a new flag "docker" in build pipeline definition
-  Added code in runner.SetupEnvironment to check for "docker:true" flag in pipeline definition and based on whether rdd-service-uri is provided ( mostly for hosted) or not ( for local cli builds) - query the RDD API service and determine RDD URI to be injected in box
- Changes to box.Run - inject a DOKER_HOST env variable in box container based on RDD URI, in case of local builds - also mount local docker socker to box
- New file rdd.go and rdd_test.go - implment logic to access RDD API service
- Added client files for rdd service
- Updated related deps in vendor.json


lake-of-dreams:
Have been tested with a "stubbed" version of RDD service on a EC2 box including all error scenarios - https://dev.wercker.com/ravi-r-singhal/rdd-fn-example/runs

Documentation/command/build.adoc
```diff
@@ -141,3 +141,10 @@ OPTIONS
   Wercker host to use for wercker reporter.
 --wercker-token::
   Wercker token to use for wercker reporter.
+--rdd-service-uri::
+  RDD API Service endpoint
```
markxnelson: probably should explain what "RDD" is?

core/options.go
```diff
@@ -710,7 +711,7 @@ func NewBuildOptions(c util.Settings, e *util.Environment) (*PipelineOptions, er
 		return nil, err
 	}
 	if pipelineOpts.RunID == "" {
-		pipelineOpts.RunID = uuid.NewRandom().String()
+		pipelineOpts.RunID = bson.NewObjectId().Hex()
```
markxnelson: what is this change for? could this cause issues?

core/box.go
```diff
@@ -35,6 +35,6 @@ type Box interface {
 	Restart() (*docker.Container, error)
 	AddService(ServiceBox)
 	Fetch(context.Context, *util.Environment) (*docker.Image, error)
-	Run(context.Context, *util.Environment) (*docker.Container, error)
+	Run(context.Context, *util.Environment, string) (*docker.Container, error)
```
markxnelson: I wonder if we should use a stronger type that just string?

rdd/rdd.go
```diff
@@ -0,0 +1,185 @@
+//   Copyright © 2018, Oracle and/or its affiliates.  All rights reserved.
```
markxnelson: replace the copyright sign with the three characters (c)

rdd/rdd.go
```diff
@@ -0,0 +1,185 @@
...
+	}
+}
+
+//verify - verify the RDD url by executing 	docker --version command
+func (rdd *RDD) verify() error {
+	rddURI := rdd.rddDetails.rddURI
+	dockerClient, err := client.NewClientWithOpts(client.WithHost(rddURI))
+	if err != nil {
+		return fmt.Errorf(`Unable to create a docker client with RDD URI: %s, Error: %s
+		`, rddURI, err.Error())
```
markxnelson: why the new line in the middle of this string?

rdd/rdd.go
```diff
@@ -0,0 +1,185 @@
...
+
+	return rdd, nil
+}
+
+//Get - Invokes RDD Service to get remote docker daemon URL by first executing a Provision()
+//request followed by polling GetStatus()
+func (rdd *RDD) Get() (string, error) {
+	rddProvRequest := &rddpb.RDDProvisionRequest{RunID: rdd.runID}
+	rddProvResponse, err := rdd.rddClient.Provision(rdd.ctx, rddProvRequest)
+
```
nigeldeakin: Delete blank line between function call and checking the error

rdd/rdd.go
```diff
@@ -0,0 +1,185 @@
...
+			}
+			log.Info(fmt.Sprintf("runID: %s, RDD Service URI: %s, RDD Provisioning status: %s", rdd.runID, rdd.rddServiceEndpoint, currentRDDState.String()))
+
+		}
+	}
+}
+
+//Delete - removes a RDD
+func (rdd *RDD) Delete() {
```
nigeldeakin: It would be more consistent to use the word "Deprovision" rather than "Delete" since that's the term used by the server API.

rdd/rdd.proto
```diff
@@ -0,0 +1,66 @@
...
+
+message ActionResponse {
+  bool success = 1;
+}
+
+message Resource {
+  string kind = 1;
+}
+
+
```
nigeldeakin: Since this whole file needs to be step in step with https://github.com/wercker/rdd/blob/master/rdd.proto, it might be helpful to add a comment to https://github.com/wercker/rdd/blob/master/rdd.proto saying this.

rdd/rdd.go
```diff
@@ -0,0 +1,185 @@
...
+	rddResponseID := rddProvResponse.GetId()
+	if rddResponseID == "" {
+		errMsg := fmt.Sprintf(errorMsgInvalidProvisioningResponse, rdd.rddServiceEndpoint, rdd.runID)
+		log.Error(errMsg)
+		return "", cli.NewExitError(errMsg, 1)
+	}
+
+	timeoutThresholdInSeconds, err := time.ParseDuration(fmt.Sprintf("%ds", rdd.rddProvisionTimeout))
+	if err != nil {
+		log.Error("Error parsing timeout value from input rdd-provision-timeout of %d, Error: %s. Default value of 300s will be used.", rdd.rddProvisionTimeout, err.Error())
```
nigeldeakin: Log a warning, not an error (log an error when you can't perform the requested operation, which is not the case here)

rdd/rdd.go
```diff
@@ -0,0 +1,185 @@
...
+	"gopkg.in/urfave/cli.v1"
+)
+
+const (
+	errorMsgFailOnProvision             = "Error invoking Provision() from rdd service at %s for runID %s, Error: %s"
+	errorMsgInvalidProvisioningResponse = "Invalid response by Provision() from rdd service at %s for runID %s, ResponseID is empty."
+	errorMsgTimeOut                     = "RDD provisioning timed out from rdd service at %s for runID %s after %d seconds."
+	errorMsgGetStatusError              = "Error provisioning RDD from rdd service at %s for runID %s. Aborting."
+	errorMsgInvalidRDDUrl               = "Invalid RDD uri returned from rdd service at %s for runID %s. Aborting."
+)
```
nigeldeakin: My preference would be to use "docker daemon" or "remote docker daemon" rather than "RDD", since it is possible the user might see this message. Suggest that all messages include the words "Cannot provision remote docker daemon" or similar. (Yes, I realise opinions differ on error messages)

Why do some of these messages use the word "aborting" and some not? You don't say what is being aborted. Suggest avoiding this word.

cmd/runner.go
```diff
@@ -522,6 +527,29 @@ func (p *Runner) SetupEnvironment(runnerCtx context.Context) (*RunnerShared, err
 		sr.Message = err.Error()
 		return shared, err
 	}
+	rddURI := ""
+	if pipeline.Docker() {
+		if p.dockerOptions.RddServiceURI != "" {
```
nigeldeakin: Is it worth adding a comment to explain what is going on here: that this is code to start a remote docker daemon, using a separate "rdd service"? Since rdd is closed-source this might be a bit mysterious.

rdd/rdd.go
```diff
@@ -0,0 +1,185 @@
...
+	}
+}
+
+//Delete - removes a RDD
+func (rdd *RDD) Delete() {
+	rddDeProvRequest := &rddpb.RDDDeprovisionRequest{Id: rdd.rddDetails.rddProvisionRequestID}
+	_, err := rdd.rddClient.Deprovision(rdd.ctx, rddDeProvRequest)
+	if err != nil {
+		errMsg := fmt.Sprintf("Error invoking Deprovision() from rdd service at %s for runID %s, Error: %s. Ignoring", rdd.rddServiceEndpoint, rdd.runID, err.Error())
+		log.Error(errMsg)
```
nigeldeakin: This should be a warning rather than an error (since this doesn't prevent the pipeline coming to an orderly end)

core/pipeline.go
```diff
@@ -68,6 +68,7 @@ type Pipeline interface {
 	DockerRepo() string
 	DockerTag() string
 	DockerMessage() string
+	Docker() bool
```
nigeldeakin: I realise that none of the other functions have documented, but I think the `Docker()` function should have a comment. (e.g. "returns whether a docker daemon is required")

Since it returns a boolean (rather than a "docker") perhaps change the name to "DockerRequired()" or some similar adjectival name.

core/config.go
```diff
@@ -188,6 +188,7 @@ type PipelineConfig struct {
 	StepsMap   map[string][]*RawStepConfig
 	Services   []*RawBoxConfig `yaml:"services"`
 	BasePath   string          `yaml:"base-path"`
+	Docker     bool            `yaml:"docker"`
```
nigeldeakin: This line is the entire user interface to this feature!

When convenient, can you add a comment to https://jira.oraclecorp.com/jira/browse/WRKR-436 (which is the JIRA for documenting this feature) to say how a user configures their wercker.yaml to request a docker daemon?

core/options.go
lake-of-dreams: This change was done because API Service developed by Nigel requires RunID as a Hex of bson.ObjectID. It will not cause any issues because while running in hosted mode - kiddie pool passes a RunId to wercker pod (and hence pipelineOpts.RunID is never empty on app.wercker or dev.wercker  )which is also in this format. I have tested that locally also it causes no issues changing it to bson.ObjectID.Hex()

Documentation/command/build.adoc
lake-of-dreams: Added

core/box.go
lake-of-dreams: RDD URI is the only field required for the box to run and all other fields returned by RDD API services are of not much use from a build execution prospective in wercker - therefore just the RDD URI is being passed..

rdd/rdd.go
lake-of-dreams: done

rdd/rdd.go
lake-of-dreams: Fixed

rdd/rdd.go
lake-of-dreams: Fixed

rdd/rdd.go
lake-of-dreams: Fixed

rdd/rdd.proto
lake-of-dreams: Added comment

rdd/rdd.go
lake-of-dreams: Fixed

rdd/rdd.go
lake-of-dreams: Fixed. Thanks for pointing out

cmd/runner.go
lake-of-dreams: Added comments

core/pipeline.go
lake-of-dreams: I have added a comment - but I am more in favor if keeping it named as Docker() returning  boolean since it is defined as "docker:true" in yaml as well and it is more useful to keep it same so as to derive a correlation between what is defined in yaml and how does this field related to that.

cmd/flags.go
```diff
@@ -50,6 +50,8 @@ var (
 		cli.IntFlag{Name: "docker-kernel-memory", Usage: "Set docker kernel memory limit in MB NOTIMPLEMENTED", Hidden: true},
 		cli.BoolFlag{Name: "docker-cleanup-image", Usage: "Remove image from the Docker when finished pushing them", Hidden: true},
 		cli.StringFlag{Name: "docker-network", Value: "", Usage: "Docker network name.", Hidden: true},
+		cli.StringFlag{Name: "rdd-service-uri", Value: "", Usage: "RDD API Service endpoint."},
```
hatchan: make these flags hidden by adding `, Hidden: true`

hatchan: The documentation shouldn't have a trailing `.`

cmd/runner.go
```diff
@@ -86,6 +87,7 @@ type Runner struct {
 	logger        *util.LogEntry
 	emitter       *core.NormalizedEmitter
 	formatter     *util.Formatter
+	rddImpl       *rdd.RDD
```
hatchan: rddImpl seems a bit weird, just `rdd` seems better to me, so `rddClient`

core/options.go
hatchan: Also, the hosted version uses a ObjectId. So this change it closer to the hosted implementation.

core/pipeline.go
```diff
@@ -68,6 +68,8 @@ type Pipeline interface {
 	DockerRepo() string
 	DockerTag() string
 	DockerMessage() string
+	//Indicates whether the build requires a Remote Docker Daemon
+	Docker() bool
```
hatchan: Start with the name of what you are describing. So something like: `Docker indicates ....`

core/pipeline.go
```diff
@@ -304,3 +306,7 @@ func (p *BasePipeline) SyncEnvironment(sessionCtx context.Context, sess *Session
 
 	return nil
 }
+
+func (p *BasePipeline) Docker() bool {
```
hatchan: Missing documentation

rdd/rdd.proto
```diff
@@ -0,0 +1,67 @@
...
+}
+message RDDProvisionResponse {
+  string id = 1;
+}
+
+message RDDStatusRequest{
+  string id = 1;
+}
+
+enum DaemonState {
```
hatchan: DaemonState was updated in the latest master version of rdd

rdd/rdd_test.go
```diff
@@ -0,0 +1,175 @@
+package rdd
+
+import (
+	"errors"
+	"fmt"
+	"reflect"
+	"testing"
+
+	ctx "context"
```
hatchan: aliasing this as ctx is very uncommon, and it conflicts with the normal convention of having the variable name `ctx`

rdd/rdd_test.go
```diff
@@ -0,0 +1,175 @@
...
+
+	context "golang.org/x/net/context"
```
hatchan: Isn't it possible to move to the `context` package?

rdd/rdd.go
```diff
@@ -0,0 +1,183 @@
...
+	errorMsgGetStatusError              = "Error provisioning Remote Docker Daemon from Remote Docker Daemon API service at %s for runID %s."
+	errorMsgInvalidRDDUrl               = "Invalid Remote Docker Daemon uri returned from Remote Docker Daemon API service service at %s for runID %s."
+)
+
+//RDD - struct containing all parameters and client for RDD access
+type RDD struct {
+	rddServiceEndpoint  string
+	rddProvisionTimeout int64
+	runID               string
+	ctx                 context.Context
```
hatchan: Do not embed the context as a field, but pass it along to the functions

rdd/rdd.go
```diff
@@ -0,0 +1,183 @@
...
+	rddDetails          *rddDetails
+}
+
+type rddDetails struct {
+	rddURI                string
+	rddProvisionRequestID string
+}
+
+//Init - initialize a RDD construct, check connection with RDD API service and create a client
+func Init(ctx context.Context, rddServiceEndpoint string, rddProvisionTimeout int64, runID string) (*RDD, error) {
```
hatchan: Use `New` as the constructor name

rdd/rddpb/generate-protobuf.sh
```diff
@@ -0,0 +1,34 @@
...
+    -I$GOPATH/src \
+    -I./vendor \
+    -I./vendor/github.com/grpc-ecosystem/grpc-gateway/third_party/googleapis"
+fi
+
+cd $LOCAL
+
+echo "Generating gRPC server, gateway, swagger, flow"
+$protoc --go_out=plugins=grpc:$ROOT/rddpb \
+        --grpc-gateway_out=logtostderr=true,request_context=true:$ROOT/rddpb \
```
hatchan: we do not need the gateway, swagger or flow in this situation. Also, could you investigate if it is possible to _only_ generate the client code? As currently the server code is also generated.

core/config.go
lake-of-dreams: Added documentation as comment in doc jira

docker/options.go
```diff
@@ -26,19 +26,21 @@ import (
...
+	CPUPeriod           int64
+	CPUQuota            int64
+	Memory              int64
+	MemoryReservation   int64
+	MemorySwap          int64
+	KernelMemory        int64
+	CleanupImage        bool
+	NetworkName         string
+	RddServiceURI       string
+	RddProvisionTimeout int64
```
hatchan: Can you change this to a `time.Duration`

cmd/flags.go
lake-of-dreams: Fixed

cmd/runner.go
lake-of-dreams: Fixed

core/pipeline.go
lake-of-dreams: Fixed

core/pipeline.go
lake-of-dreams: Added

rdd/rdd_test.go
lake-of-dreams: Fixed

rdd/rdd_test.go
lake-of-dreams: not sure what you mean. https://github.com/wercker/rdd/blob/master/rddpb/rdd.pb.go uses context "golang.org/x/net/context"

rdd/rdd.go
lake-of-dreams: Fixed

rdd/rdd.go
lake-of-dreams: Done

docker/options.go
lake-of-dreams: I think it is more appropriate to take it in as a primitive integer - in rdd.go I parse this to check if it can be converted to seconds duration and if not, I use a default

rdd/rddpb/generate-protobuf.sh
lake-of-dreams: TBD..

docker/options.go
hatchan: I don't understand why you want to keep it a primitive, as you say. In both cases we're going to have to do some parsing. So then I would like the parts that we as dev/ops are touching to be as easy to use.

For example if the input is `780`, I have no idea how much it will be, whereas `7m` is way more clear.

lake-of-dreams: wercker/wercker was using a dependency github.com/wercker/codegangsta-cli which was a stale and outdated fork from github.com/urfave/cli ( in a branch "hidden_flags" which was not updated since last 3 years - last commit by @hatchan 3 years back). Moreover the flag "DurationFlag" was not correctly implemented as it did not implement "ishidden" method and therefore it was not possible to implement this comment.


 Surprisingly kiddie-pool was using github.com/urfave/cli only. Since it seems the intention of the fork was to introduce feature of hidden flags and now github.com/urfave/cli itself contains functionality of hidden flags. Also we should be consistent in our dep usage across modules to the best extent possible. It was also not fruitful to keep updating the stale fork and instead I have changed wercker CLI code to start using latest github.com/urfave/cli as deep instead of github.com/wercker/codegangsta-cli.

lake-of-dreams:
https://github.com/wercker/wercker/pull/421#discussion_r194346111

PR #423, opened 2018-06-11
docker-push fixes: display messages in web, cease defaulting to wcr.io

nigeldeakin:
This PR contains two sets of changes related to `internal/docker-push`

https://github.com/wercker/wercker/commit/d80c73f1abf830b2df17a514125c116b803b9999
fixes a bug which prevented any of the information messages logged by `InferRegistryAndRepository` from being displayed in the web UI. This is because the messages were simple log messages. These are now logged using the Emitter which displays them in the web UI.

https://github.com/wercker/wercker/commit/b7fc2f7ff5eab3fe02b34b4a7d73f437d9f3bdce
changes the behaviour when the repository is not specified. Previously this defaulted to `ipelineOptions.WerckerContainerRegistry` (which is currently https://wcr.io/v2 in production and https://registry-alpha.oracledx.com/v2 in staging). Now it simply fails the step with a message "repository not set"

This PR is required to fix https://jira.oraclecorp.com/jira/browse/WRKR-153


docker/docker.go
```diff
@@ -648,10 +648,15 @@ func (s *DockerPushStep) buildAutherOpts(env *util.Environment) (dockerauth.Chec
 	return opts, nil
 }
 
+// log logs the specified message using the specified emitter
+func log(e *core.NormalizedEmitter, message string) {
```
lake-of-dreams: should this be renamed to a more genric function name such as "emit" and moved to a shared util - I think this can easily be reused at several other places

nigeldeakin: I've renamed it to `emit` as suggested (since that's what it does). This is in a file called `docker.go` which is probably the best place for a shared utility that can be used by other functions in this package. (At some point in the future we should move all the code specific to docker-push into a separate file `docker-push.go`, leaving `docker.go` for shared utilities.)

nigeldeakin:
I've removed `debug.test` from the commit. Well spotted.

nigeldeakin:
Re-tested in staging.

1. `docker-push` without repository set (to test the main use case this issue refers to)

Production (pushes to wcr.io successfully without any messages):
https://app.wercker.com/nigeldeakin/wercker-tests/runs/test-pipeline/5b28e00b076d4400011e6306?step=5b28e0310908c400010ca798

Staging (fails with repo not specified):
https://dev.wercker.com/nigeldeakin/wercker-tests/runs/test-pipeline/5b28ddca28198600013abbbc?step=5b28de1978eb58000118b51a

2. `docker-push` with conflicting registry and repository, with a non-existent registry (to test whether those info messages are displayed)

Production: (does not mention the conflicting registry and repository and fails due to the non-existent registry)
https://app.wercker.com/nigeldeakin/wercker-tests/runs/test-pipeline2/5b28e00b076d4400011e630a?step=5b28e03d4fea0c00011eb570

Staging: (reports the conflicting registry and repository and which was used, and then fails due to the non-existent registry)
https://dev.wercker.com/nigeldeakin/wercker-tests/runs/test-pipeline2/5b28ddca28198600013abbc0?step=5b28de21fdd92e0001ae1d3d

nigeldeakin:
Ready for re-review

PR #424, opened 2018-06-13
Fix for WRKR-387 (docker-build step reports success even when failure)

nigeldeakin:
This PR fixes https://jira.oraclecorp.com/jira/browse/WRKR-387

If the docker build fails due to an error in the Dockerfile then an error is reported in the output stream but the Docker API call itself returns without an error.

The code that parsed the output stream (`EmitStatus`) already spotted that an error had been reported and returned an error. However the code that called it (`docker_build.go`) ignored the error. This is now been fixed: if an error is encountered in the output stream then this causes the step to fail (which is what `docker_push` already does). 

I have also changed `EmitStatus` so that the error message (and a following \n) is logged using the emitter. This means that the error message is also displayed in the black step output window  in wercker.com (which is where the user would expect it) as well as in the italic "reason the step failed" line above it.

Tested in staging: see details of before and after testing [here](https://jira.oraclecorp.com/jira/browse/WRKR-387?focusedCommentId=30134442&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-30134442)


docker/status_handler.go
```diff
@@ -43,7 +43,7 @@ type JSONMessageProcessor struct {
 // message and previous messages.
 func (s *JSONMessageProcessor) ProcessJSONMessage(m *jsonmessage.JSONMessage) (string, error) {
 	if m.Error != nil {
-		return "", m.Error
+		return m.ErrorMessage + "\n", m.Error
 	}
```
agnihotrigaurav: I suppose we should avoid inserting "\n" in error messages. Could be issue while loading logs in ELK.

docker/status_handler.go
```diff
@@ -43,7 +43,7 @@ type JSONMessageProcessor struct {
 // message and previous messages.
 func (s *JSONMessageProcessor) ProcessJSONMessage(m *jsonmessage.JSONMessage) (string, error) {
 	if m.Error != nil {
-		return "", m.Error
+		return m.ErrorMessage + "\n", m.Error
```
agnihotrigaurav: I suppose we should avoid inserting "\n" in error messages. Could be issue while loading logs in ELK.

docker/events.go
```diff
@@ -37,14 +37,13 @@ func EmitStatus(e *core.NormalizedEmitter, r io.Reader, options *core.PipelineOp
 		}
 
 		line, err := s.ProcessJSONMessage(&m)
-		if err != nil {
```
lake-of-dreams: Not sure why it was moved down, if there was an error in ProcessJSONMessage - call should return from there itself - or if we only want to return after Emit() - then Emit should also include message from the err

docker/status_handler.go
nigeldeakin: This affects only the string passed to the emitter. In normal operation the \n is extracted from the incoming JSON and returned. However in the case of an error there is no \n and so one is appended here. 

I could remove this, but it would mean that the following line of output is on the same line.

docker/events.go
nigeldeakin: I moved it down so that the string returned from `ProcessJSONMessage(&m)` was sent to the emitter even if an error had occurred.

See the change to `docker/status_handler.go `. This now returns the error message as he first return value even if an error is returned as the second return value.

These two changes combined mean that if an error occurs, the error message is sent to the emitter. This displays it in the black output window. 

If I omitted this change the error message (which comes from the docker build output) would not appear in the black output window. It would still appear at the top of the window in italics.

nigeldeakin:
I'll write a unit test.

nigeldeakin:
Thanks for the comments. I have made the following changes:
- I have modified the actual bugfix so that the only code changed is `EmitStatus` in `events.go`. If `ProcessJSONMessage` (which is now unchanged) returns an error this is now output by the emitter.
- I have refactored `DockerBuildStep.Execute` in `docker_build.go`  to make it easier to test in a unit test. Essentially this is divided into two stages, each in a separate function. The first is `buildTarfile` which creates the tarfile for sending to docker, and the second is `buildImage`which sends the tarfile to docker and handles the output.
- I have created two new unit tests `TestBuildFailure` and `TestBuildSuccess` `docker_build_test.go` which verify that an error when processing the Dockerfile will cause  `DockerBuildStep.Execute` to return an error.

Please re-review

nigeldeakin:
Tested in staging: 
https://dev.wercker.com/nigeldeakin/docker-build-golang/runs/build/5b28cd5f28198600013abb0b?step=5b28cd6378eb58000118b258

This is what it looks like in production (without this fix)
https://app.wercker.com/nigeldeakin/docker-build-golang/runs/build/5b28cd6098f23800012475a2?step=5b28cd7b0908c400010ba1bb

hatchan:
(btw wasn't able to view the build on app, received a 401)

lake-of-dreams:
Please add 

- Fix wercker --help for subcommands(#432)

under you changelog version also (changes from https://github.com/wercker/wercker/pull/433 which as per Benno should be bundeled with your changes)

PR #428, opened 2018-06-16
Collect artifacts from steps even if they fail (WRKR-566)

markxnelson:
This PR addresses requests from two users to preserve the contents of the reports artifact directory even if a step fails. 

This is needed because users are storing logs, screenshots, test results, etc., in the reports directory.  These are then easily accessible from the web UI after the job is complete.  However, the current behavior only saves the artifacts if the step passed.  Users actually want to be able to get the reports even (and in fact especially) when the step failed. 

This change will store any report artifacts for a step whether that step passed or failed.  Verified with a local build (and --debug) and also deployed to staging and verified that report artifacts from a failing step are uploaded to s3.


cmd/runner.go
```diff
@@ -778,6 +770,14 @@ func (p *Runner) RunStep(ctx context.Context, shared *RunnerShared, step core.St
 		sr.Artifact = artifact
 	}
 
+	// This is the error from the step.Execute above
+	if err != nil {
```
hatchan: Not sure if the `err` object will be set in the if statement above :thinking:

hatchan: Guess it doesn't but it would be quite easy for a developer to make the mistake of reusing the error object.

PR #436, opened 2018-06-21
Remove call to docker-check-access CheckAccess method

markxnelson:
Currently the CLI calls code in docker-check-access before it tries to access a docker registry. This code implements a variety of special cases that try to determine if a user has access to a registry. It works by making a REST call over HTTP directly to the registry. It does not run on the daemon, and so does not pick up any configuration on the daemon, like certificates, insecure_registries, etc., that may actually be important. We feel that this code should be deleted. Not only does it not work, but also it is going to be a real pain to maintain it since we would have to be constantly implementing special cases (like artifactory docker registry support for example) and keeping them up to date. And there is really no good reason for wercker checking access when the very next command (usually a docker push) is going to do that anyway.

This PR removes the calls to the CheckAccess code, and is related to https://github.com/wercker/docker-check-access/pull/10 in wercker/docker-check-access which removes its use of that method.


gilbode:
There is a bunch more code potentially obsolete by removing the use of docker-check-access.  I think we should remove it all.

nigeldeakin:
We've had that discussion and we agreed we were happy to proceed.

PR #438, opened 2018-06-22
escape dollar revert

agnihotrigaurav:
Reverting escape dollar change from master.


nigeldeakin:
Why didn't you simply revert the commit (using `git revert`)?

Then add a further commit to update the changelog. This should show the old, bad, release and the new, good, release, so we can keep track of history. We shouldn't simply obliterate mention of the old, bad, release.

PR #439, opened 2018-06-22
Bihaber wrkr 611 540 581

bhabermaas:
WRKR-611 Add config-id parameter to runner start
WRKR-540 When production, allow :latest image tag as preferred, if not found then use most recent master. When not production use most recent master.  
WRKR-540 Add image-name parameter to runner start to allow image override for development 
WRKR-581 Default --group from --name


bhabermaas:
The pull is no longer valid based on design changes.

PR #442, opened 2018-06-27
Run pipeline in remote docker daemon (initial version)

nigeldeakin:
If the user specifies that they require a remote docker daemon (using the pipeline property `docker: true`) then the pipeline container will be run in that daemon rather than the restricted, shared, daemon. 

The most immediate benefit of this change is it allows the pipeline container to connect to any containers that it starts using `docker run` (or equivalent). The `docker run` parameter `--network=$DOCKER_NETWORK_NAME` must be specified when starting any additional containers to add them to the same network as is used by the pipeline container.

(Obviously we need to do this automatically and remove the need to specify this)

This is not yet ready for merging: further review is required and both unit and end-to-end tests are needed.

Here is a summary of the main changes so far:
* If `wercker.yml` specifies `docker: true` then a RDD is provisioned as before and URI of the daemon saved in the docker setting `dockerOptions.Host`. This means that when the "box" container and any service containers are started they are started in the RDD daemon rather than in the default, shared daemon.
* A new environment variable `DOCKER_NETWORK_NAME` is set to the name of the docker network that is created. This must currently be specified when creating any other containers.
* Prior to this change the step directories were made available to the box container using mounts (binds). However this is not possible if the daemon is running on a different machine, so if a RDD is being used the step directory is copied to the remote docker daemon. This involves creating a tarball and calling the `CopyToContainer` API call.
* If the CLI is being run locally and `wercker.yml` specifies `docker: true` then the local daemon is used (as before) and a new daemon is not started. I have maintained the prior behaviour which mounts the docker socket inside the container: we must be careful do this only if the CLI is being run locally.  In this case the step directories are copied to the local daemon via a tarball. I did this to make it easier to test but we need to change this to use a mount as this is much faster.
* I have split the existing Box.bind() function into two separate functions, `mounts` (which mounts the step directories) and `vols` (which exposes the docker socket) to simplify the logic that controls when each is used.
* I have been testing this with https://github.com/nigeldeakin/direct-docker-golang/ (which will one day become an official example) and my own test repo https://github.com/nigeldeakin/wercker-tests . Obviously lots more tests are needed


docker/box.go
```diff
@@ -378,7 +428,25 @@ func (b *DockerBox) Run(ctx context.Context, env *util.Environment, rddURI strin
...
+		// Not using a RDD
+		// Obtain the binds necessary to mount the directories under HostPath (which is a pipeline option),
+		b1, err := b.mounts(env)
+		if err != nil {
+			return nil, err
+		}
+		binds = append(binds, b1...)
+	}
+	if b.options.EnableVolumes {
+		// Obtain the binds necessary to mount the volumes defined in Volumes (which is a pipeline option)
```
agnihotrigaurav: need to check various volume mounting options also will it be possible to mount same on rdd?

docker/box.go
```diff
@@ -445,6 +513,14 @@ func (b *DockerBox) Run(ctx context.Context, env *util.Environment, rddURI strin
 		return nil, err
 	}
 
+	if rddURI != "" {
```
agnihotrigaurav: Local case should be checked as well.

docker/box.go
```diff
@@ -166,27 +170,70 @@ func (b *DockerBox) binds(env *util.Environment) ([]string, error) {
...
+	if err != nil {
+		return err
+	}
+
+	entries, err := ioutil.ReadDir(b.options.HostPath())
+	if err != nil {
+		return err
+	}
+	for _, entry := range entries {
+		if entry.IsDir() || entry.Mode()&os.ModeSymlink == os.ModeSymlink {
```
lake-of-dreams: os.ModeSymlink == os.ModeSymlink ??

docker/box.go
```diff
@@ -166,27 +170,70 @@ func (b *DockerBox) binds(env *util.Environment) ([]string, error) {
...
+			if err != nil {
+				return err
+			}
+			// now create a reader on the tarball
+			tarFile, err := os.Open(tarfileName)
+			tarReader := bufio.NewReader(tarFile)
+
+			// copy the tarball to the container, where it will be untarred in the specified location
+			err = client.CopyToContainer(ctx, containerID, destDirName, tarReader, types.CopyToContainerOptions{})
+			if err != nil {
```
lake-of-dreams: This should be tested thoroughly - e.g. can we mount a file?

docker/box.go
```diff
@@ -378,7 +430,25 @@ func (b *DockerBox) Run(ctx context.Context, env *util.Environment, rddURI strin
 		ports = exposedPorts(b.config.Ports)
 	}
 
-	binds, err := b.binds(env)
+	binds := []string{}
+	if rddURI == "" {
+		// Not using a RDD
+		// Obtain the binds necessary to mount the directories under HostPath (which is a pipeline option),
+		b1, err := b.mounts(env)
```
lake-of-dreams: a better var name than b1

lake-of-dreams:
One more scenario that we need to handle is how to handle a situation when a user starts a container with a volume and path that the user mounts exists on box container

`docker run -v `pwd`:/go/src/github.com/fnproject/fn --rm fnproject/swagger:0.0.1 /go/src/github.com/fnproject/fn/docs/swagger.yml`

/go/src/github.com/fnproject/fn exists on the box container but since path /go/src/github.com/fnproject/fn will not be on rdd host - it will not work

docker/box.go
nigeldeakin: Changed to `if rddURI == "" || b.dockerOptions.RddServiceURI == "" {`

docker/box.go
nigeldeakin: I haven't changed anything here. This line is in the current version:
https://github.com/wercker/wercker/blob/54c57f3c6637a9b4637e85eb311ea5f27a664fec/docker/box.go#L157
I think `entry.Mode()&os.ModeSymlink == os.ModeSymlink` tells you whether it is a symbolic link. I think you could also use `entry.Mode()&os.ModeSymlink != 0` but I don't want to change existing logic.

nigeldeakin:
Extended the existing test for `--direct-mount` with a variant that uses a RDD

nigeldeakin:
(Pasted from Slack)
- I'm trying to create pipeline to build docker images for OracleFMWInfrastructure (https://github.com/oracle/docker-images/blob/master/OracleFMWInfrastructure/dockerfiles/12.2.1.3/Dockerfile).
- Orahub project : https://orahub.oraclecorp.com/scqs/wercker-fmw-docker-images/tree/master

- Here Dockerfile  uses "FROM oracle/serverjre:8" and hence I'm building the image oracle/serverjre:8 in the same pipeline which is success before building FMW Infra image.

- With Wercker UI (app.wercker.com) (Run details: https://app.wercker.com/paas-rm/wercker-fmw-docker-images/runs/build/5b4621c887e17d0001aa2d25?step=5b4621e263c37f0001b691be)   the pipeline shows success but does not build the FMW Infra image and just shows "Step 1/12 : FROM oracle/serverjre:8".

- With wercker cli able to get the below error details:
=====================
Step 1/12 : FROM oracle/serverjre:8
repository oracle/serverjre not found: does not exist or no pull access
--> repository oracle/serverjre not found: does not exist or no pull access
--> Step failed: Building FMW Infra docker images repository oracle/serverjre not found: does not exist or no pull access 60.57s
--> Pipeline failed: 183.44s
FATAL Step failed: Building FMW Infra docker images
========================

nigeldeakin:
Extended and refactored the existing test for `--enable-volumes` to test the mounting of volumes in both the box container and any service containers, and when a RDD is being used and when it is not.

nigeldeakin:
Updated the description with details of what tests I have added

docker/box.go
```diff
@@ -378,7 +451,26 @@ func (b *DockerBox) Run(ctx context.Context, env *util.Environment, rddURI strin
...
+		extraBinds, err := b.mounts(env)
+		if err != nil {
+			return nil, err
+		}
+		binds = append(binds, extraBinds...)
+		stepsMounted = true
+	}
+	if b.options.EnableVolumes {
+		// Obtain the binds necessary to mount the volumes defined in Volumes (which is a pipeline option).
+		b2, err := b.vols(env)
```
lake-of-dreams: there should be a better variable name than b2

nigeldeakin: Changed to `volBinds`since its value is set to `b.vols(env)` and is appended to a slice called `binds`

docker/box.go
```diff
@@ -166,27 +172,89 @@ func (b *DockerBox) binds(env *util.Environment) ([]string, error) {
...
+	for i := 0; i < len(b.volumes); i += 2 {
+		binds = append(binds, fmt.Sprintf("%s:%s:rw", b.volumes[i], b.volumes[i+1]))
+	}
+	return binds, nil
+}
+
+// copyStepsToContainer() copies the directories under HostPath (which is a pipeline option) to the pipeline container.
+// This is equivalent to calling mounts() to mount the same directories, and must be used when the daemon is remote and binds are not possible.
+func (b *DockerBox) copyStepsToContainer(ctx context.Context, env *util.Environment, containerID string) error {
+
```
hatchan: can you remove these leading empty lines after the func deceleration?

docker/box.go
```diff
@@ -166,27 +172,89 @@ func (b *DockerBox) binds(env *util.Environment) ([]string, error) {
...
-		for _, vol := range vols {
-			if strings.Contains(vol, ":") {
-				pair := strings.SplitN(vol, ":", 2)
-				interpolatedVols = append(interpolatedVols, env.Interpolate(pair[0]))
-				interpolatedVols = append(interpolatedVols, env.Interpolate(pair[1]))
-			} else {
-				interpolatedVols = append(interpolatedVols, env.Interpolate(vol))
-				interpolatedVols = append(interpolatedVols, env.Interpolate(vol))
-			}
+// vols() returns the binds necessary to mount the volumes defined in Volumes (which is a pipeline option)
```
hatchan: no need to have the `()` there

docker/box.go
```diff
@@ -143,7 +147,9 @@ func (b *DockerBox) GetID() string {
 	return ""
 }
 
-func (b *DockerBox) binds(env *util.Environment) ([]string, error) {
+// mounts() returns the binds necessary to mount the directories under HostPath (which is a pipeline option).
```
hatchan: no need to have the `()` there

docker/box.go
```diff
@@ -166,27 +172,89 @@ func (b *DockerBox) binds(env *util.Environment) ([]string, error) {
...
-			binds = append(binds, fmt.Sprintf("%s:%s:rw", b.volumes[i], b.volumes[i+1]))
+	}
+	b.volumes = interpolatedVols
+	for i := 0; i < len(b.volumes); i += 2 {
+		binds = append(binds, fmt.Sprintf("%s:%s:rw", b.volumes[i], b.volumes[i+1]))
+	}
+	return binds, nil
+}
+
+// copyStepsToContainer() copies the directories under HostPath (which is a pipeline option) to the pipeline container.
```
hatchan: no need to have the `()` there

hatchan:
Oh update the changelog in the PR

docker/box.go
```diff
@@ -166,27 +172,89 @@ func (b *DockerBox) binds(env *util.Environment) ([]string, error) {
...
+	for _, entry := range entries {
+		if entry.IsDir() || entry.Mode()&os.ModeSymlink == os.ModeSymlink {
+
+			sourceDirName := b.options.HostPath(entry.Name())
+			sourceDirName = strings.TrimRight(sourceDirName, "/")
+			destDirName := b.options.MntPath("")
+
+			// if the path is a symlink then follow it before calling TarPathWithRoot (since TarPathWithRoot ignores symlinks)
+			if entry.Mode()&os.ModeSymlink == os.ModeSymlink {
+				sourceDirName, err = filepath.EvalSymlinks(sourceDirName)
```
markxnelson: are we sure we want to follow symlinks? not just copy them as symlinks?

tests/projects/direct-mount-test/test.sh
```diff
@@ -0,0 +1,71 @@
...
+if [ -z ${testsDir} ]; then testsDir=$PWD/..; fi
+
+# Test direct-mount in the non-RDD case by running the test-direct-mount-normal pipeline
+testDirectMountNormal () {
+  testName=direct-mount-test-normal
+  testDir=$testsDir/direct-mount-test
+  printf "testing %s... " "$testName"
+  
+  testFile=${testDir}/testfile-normal
+  > "$testFile"
```
markxnelson: is this redirect meant to be on its own line, or is this just a diff glitch?

tests/projects/direct-mount-test/test.sh
```diff
@@ -0,0 +1,71 @@
...
+}
+
+# Test direct-mount in the RDD case by running the test-direct-mount-rdd pipeline
+testDirectMountRDD () {
+  testName=direct-mount-test-rdd
+  testDir=$testsDir/direct-mount-test
+  printf "testing %s... " "$testName"
+
+  testFile=${testDir}/testfile-rdd
+  > "$testFile"
```
markxnelson: redirect on its own line?

tests/projects/rdd/wercker.yml
```diff
@@ -0,0 +1,62 @@
...
+                else
+                    echo "Test failed: container gave unexpected response: " $curlOutput
+                    echo "The expected response was: " $expected
+                    exit 1
+                fi   
+            else 
+                echo "Test failed: container did not respond"
+                exit 1 
+            fi           
+    - script:
```
markxnelson: after step?

util/util.go
```diff
@@ -23,6 +23,7 @@ import (
 	"net/http"
 	"os"
 	"path"
```
markxnelson: are you importing the same thing twice here, and giving it a second name?

util/util.go
```diff
@@ -289,6 +292,69 @@ func TarPath(writer io.Writer, root string) error {
...
+// in the tarball, with the files under "root" placed under it.
+// Directory names are written to the tarfile (including empty directories).
+// Symbolic links are ignored (walk doesn't follow symbolic links anyway).
+func TarPathWithRoot(writer io.Writer, root string, tarRoot string) error {
+	tw := tar.NewWriter(writer)
+	defer tw.Close()
+
+	walkFn := func(path string, info os.FileInfo, err error) error {
+
+		if info.Mode()&os.ModeSymlink == os.ModeSymlink {
```
markxnelson: i am curious to understand why we think ignoring symlinks is safe

docker/box.go
hatchan: Good catch, I agree that we probably want to copy the symlinks, and not copy the contents.

docker/box.go
nigeldeakin: Done

docker/box.go
nigeldeakin: Agreed. Done.

docker/box.go
nigeldeakin: Agreed. Done.

docker/box.go
nigeldeakin: Agreed. Done.

docker/box.go
nigeldeakin: In this loop we are iterating over the contents of the `build` directory. This is what it typically contains:
```
lrwxr-xr-x  1 ndeakin  staff   80 19 Jul 10:05 cache -> /Users/ndeakin/go/src/github.com/nigeldeakin/direct-docker-golang/.wercker/cache
drwxr-xr-x  3 ndeakin  staff   96 19 Jul 10:05 script-11d3760b-683d-47c1-add4-beed3daac470
drwxr-xr-x  3 ndeakin  staff   96 19 Jul 10:05 script-309cb98f-53e2-4fa9-a09a-c8c1c1b47ef3
drwxr-xr-x  3 ndeakin  staff   96 19 Jul 10:05 script-6a4dcba5-edd8-4399-b9cb-af1c34b6b260
drwxr-xr-x  3 ndeakin  staff   96 19 Jul 10:05 script-724e8dae-0de5-407a-89ea-1f0eee74acc0
drwxr-xr-x  3 ndeakin  staff   96 19 Jul 10:05 script-737adf9e-3189-4efd-b725-4ce21a6fbb9c
drwxr-xr-x  3 ndeakin  staff   96 19 Jul 10:05 script-a698cb61-e53f-44ab-bd67-9beb9dcce491
drwxr-xr-x  3 ndeakin  staff   96 19 Jul 10:05 script-bd625384-fcb9-4fec-b8d5-e1359e47d43c
drwxr-xr-x  3 ndeakin  staff   96 19 Jul 10:05 script-d64dc487-7bb6-4779-9b3a-61051bde7cbf
drwxr-xr-x  3 ndeakin  staff   96 19 Jul 10:05 script-fbf76cac-66cf-435d-b159-548547b0239f
lrwxr-xr-x  1 ndeakin  staff  104 19 Jul 10:05 source -> /Users/ndeakin/go/src/github.com/nigeldeakin/direct-docker-golang/.wercker/projects/direct-docker-golang
drwxr-xr-x  6 ndeakin  staff  192 19 Jul 10:05 wercker-init-ccb064fe-4406-4406-958b-a0a8841ba121
```
These are the cache, source and steps, all of which need to be tared up and sent to the remote daemon (normally they would be simply mounted).
 
`cache` and `source` are symbolic links. This is why we need to follow the link before calling `util.TarPathWithRoot`.

In addition, note that the symbolic link `cache` is to a directory `direct-docker-golang` (which is the name of the source repo). However we want the root directory in the tarfile to be `cache` (the name of the symlink). This is why we need to additionally save the name of the symbolic link in the `tarRoot` variable and pass it to `util.TarPathWithRoot`, which then uses that name as the root directory in the tarfile.

Note that the logic for handling symlinks below this level is handled by `util.TarPathWithRoot`. This uses `filepath.Walk` to walk the tree, and this does not follow symlinks. 

I am separately reviewing whether `util.TarPathWithRoot` needs to handle symlinks differently.

tests/projects/rdd/wercker.yml
nigeldeakin: Sure. Done.

util/util.go
nigeldeakin: Yes. I did this because in an earlier version of these changes I was sharing code with the existing `TarPath` function which already had a variable called `path`. This meant I could not use the `path` package without renaming the existing variable. I decided that to make the job of reviewing this PR easier (and for no other reason) I decided to rename the package instead.

However I subsequently decided to keep the two functions separate, which means I might as well rename the `path` variable rather than the import. Which I have now done.

util/util.go
nigeldeakin: If you had wondered why I had been taking so much interest in that symlink issue you discovered earlier this week, it's because of this very line.

I intend to change this (as with so many things the thing that takes the most time is writing the tests)

PR #443, opened 2018-06-28
Bihaber wrkr540 - Select runner image based on production vs development

bhabermaas:
1. Changes to accept image tag 'latest' when running in production (app.wercker.com).  When latest is not present in the remote repository it will select the most recent master.  When 'latest' is present, the associated image tag containing master and commit-id is used. This is done so the local repository will identify the commit unless the 'latest' image is pulled manually. 
2. Added --image-name parameter to runner start to bypass the normal image selection process. This parameter is hidden from help and provided for developer use. 
3. Added debugging information for images found in the local/remote repository
4. Remove 'external' from runner messages.
5. Selection of the wercker site no longer uses base-url but is replaced by --using=prod|dev with the default being prod.  The use of base-url was not working for more than a single subcommand (added support to differentiate between production/development for the configure command.
6. Fixed wrong parameters appearing in runner start help.


cmd/main.go
```diff
@@ -384,23 +384,23 @@ var (
 	runnerCommand = cli.Command{
 		Name:      "runner",
 		ShortName: "run",
-		Usage:     "manage external pipeline runners",
+		Usage:     "unmanaged pipeline runners",
 		Subcommands: []cli.Command{
 			{
 				Name:  "start",
-				Usage: "start external runner(s) - This feature is a paid option.",
+				Usage: "start runner(s) - This feature is a paid option.",
```
mgianatagh: Remove the paid option text

cmd/main.go
```diff
@@ -384,23 +384,23 @@ var (
...
 						params.RunDockerController(false)
 					}
 				},
-				Flags: FlagsFor(ExternalRunnerStartFlagSet, GlobalFlagSet),
+				Flags: FlagsFor(ExternalRunnerStartFlagSet),
 			},
 			{
 				Name:  "stop",
-				Usage: "stop external runner(s) - This feature is a paid option.",
+				Usage: "stop runner(s) - This feature is a paid option.",
```
mgianatagh: Remove the paid option text

cmd/main.go
```diff
@@ -413,7 +413,7 @@ var (
 			},
 			{
 				Name:  "status",
-				Usage: "display the status of started external runner(s) - This feature is a paid option.",
+				Usage: "display the status of started runner(s) - This feature is a paid option.",
```
mgianatagh: Remove the paid option text

cmd/main.go
```diff
@@ -425,15 +425,15 @@ var (
 			},
 			{
 				Name:  "configure",
-				Usage: "setup Docker configuration for external runner operation - This feature is a paid option.",
+				Usage: "setup Docker configuration for runner operation - This feature is a paid option.",
```
mgianatagh: Remove the paid option text

mgianatagh:
I feel we should confirm with @markxnelson about removing the paid text now.  I figure while Bill is removing the "external" text might as well remove the "paid" text too.

lfeigen:
Well, at this point, it's not *not* a paid option is it?  Somebody has to enable the feature flag, no?

mgianatagh:
Correct, the feature flag has to be enabled in the database in order to use runners

lfeigen:
So, I think saying "paid" is not such a bad term.  We could say that it is a "future feature", but that's not so cool either.  The only people who might be annoyed by seeing "paid feature" are the people who are paying and don't know the feature.

Ideas?

mgianatagh:
Maybe we should hide the runner keywords

bhabermaas:
The paid and external runner changes, if not accepted, will result in this pull request being cancelled.

PR #444, opened 2018-06-28
WRKR-611 revised to default groupname or handle groupname@organization

bhabermaas:
For unmanaged runners: this change defaults group name to the name parameter when group is not specified. It also checks if the group parameter includes an organization (groupname@organization) and when the organization is present the --apps and --workflows parameters are disallowed.  The --orgs parameter is not excepted to allow existing operation. Since server-side filter configuration is not yet implemented in Dispatch this change is benign and does not affect any functionality.


external/controller.go
```diff
@@ -181,6 +181,19 @@ func (cp *RunnerParams) RunDockerController(statusOnly bool) {
...
 	}
 
+	// Validate the group name. If omitted entirely then default to the base name. When present
+	// check for an org specification and disallow spps/workspaces parameters. A simple group name
+	// without the organization is passed along and does not invoke server-side filtering.
+	if cp.GroupName == "" {
+		cp.GroupName = cp.Basename
+	} else {
+		if strings.IndexByte(cp.GroupName, '@') != -1 {
+			if cp.AppNames != "" || cp.Workflows != "" {
```
mgianatagh: Should this also check that --orgs and --all were not specified?

bhabermaas: It was considered unnecessary but for consistency I'll add those.

PR #445, opened 2018-07-03
Use auth token when interacting w/ step registry for private steps.

gilbode:
If you try to use a private step either locally w/ the CLI or in the web UI, you get an authentication error.

https://app.wercker.com/gilbode/test-private-step/runs/test/5b3adfd1a5de59000101c4a7?step=5b3adfd1a5de59000101c4ab

This passes the auth token to the steps registry.  Works on the CLI, but not really sure how/if tokens are used by KP.


gsserge:
Reviewed, fixed tests and tested it with my private step on staging. Someone else should probably review this PR as well.

The auth token is passed to sentcli, so it's no different from running the cli locally.

core/step.go
```diff
@@ -356,20 +356,22 @@ func (s *ExternalStep) Fetch() (string, error) {
 
 	if !stepExists {
 		// If we don't have a url already
+
+		var client api.StepRegistry
+		// TODO(termie): probably don't need these in global options?
```
gmcadoo: Do we need to leave the TODO?  I think termie is not even with us any more is he?

core/step.go
```diff
@@ -356,20 +356,22 @@ func (s *ExternalStep) Fetch() (string, error) {
...
+		if s.options.GlobalOptions.StepRegistryURL == "" {
+			apiOptions := api.APIOptions{
+				BaseURL: s.options.GlobalOptions.BaseURL,
+			}
+			// NOTE(kokaz): this client doesn't contain any auth token
```
gmcadoo: Is this comment correct?  I thought this was all about doing the auth token?

core/step.go
markxnelson: I am ok with removing it

core/step.go
markxnelson: see the else :)

markxnelson: if the step registry is not empty, then the auth token is added

PR #446, opened 2018-07-05
Pass the network name to the docker-build step

hatchan:


docker/docker_build.go
```diff
@@ -225,6 +225,13 @@ func (s *DockerBuildStep) buildImage(ctx context.Context, sess *core.Session, ta
 
 	s.logger.Debugln("Build image")
 
+	// Note: This is a little hack; if a network was not passed through a flag,
+	//       then s.options.DockerNetworkName will contain the generated name.
+	networkName := s.dockerOptions.NetworkName
+	if networkName == "" {
+		networkName = s.options.DockerNetworkName
```
agnihotrigaurav: I think calling GetDockerNetworkName method would make is consistent with the way its used in services and docker run.  Otherwise chances are this is missed while refactoring docker networking code

hatchan: There is no DockerBox available in the step and creating that object just for the sake of getting the network name, seems like a bigger hack than this.

docker/docker_build.go
```diff
@@ -225,6 +225,13 @@ func (s *DockerBuildStep) buildImage(ctx context.Context, sess *core.Session, ta
...
+	}
+
```
nigeldeakin: Are you sure? I think this is just the value specified on the command line. I don't think this is set to the generated name.

nigeldeakin: Oh, forget that. I see now.

PR #454, opened 2018-07-24
Wrap errors for docker box

jwells131313:
This wraps most of the errors in box.go in order to help us triage an intermittent error a customer is having (and which we have also reproduced independently).  Not to mention, it'll help in all future triages.


gmcadoo:
FYI, @bhabermaas  is updating this as well, you might need to coordinate with him.

PR #457, opened 2018-07-25
This error condition was too late.  Now it is just in time

jwells131313:


jwells131313:
I wonder if even if there is an error the shared.box should get cleaned.  In that case the fix would be different.  Like, if shared != nil && shared.box != nil {...} ?

jwells131313:
I actually updated the fix to do the cleanup even when it fails in some ways.  It's safer

PR #458, opened 2018-07-25
Docker build auth yaml

nigeldeakin:
This PR extends the existing `internal/docker-build` step to allow the user to specify username/password for one or more registries used by the Dockerfile. This information is specified in YAML, not JSON, for consistency with the rest of `wercker.yml`.
```
    - internal/docker-build: 
        dockerfile: Dockerfile
        image-name: my-new-image
        auth:
        - registry: docker.example.com
          username: janedoe
          password: hunter2
        - username: mobydock
          password: conta1n3rize14
```
There is a new property `auth` which should be set to an array of YAML objects, each of which has the fields `registry`, `username` and `password`. If `registry` is omitted Docker Hub is assumed.

This is the format suggested by @hatchan 

Internally, when `wercker.yml` is parsed, any property which is not a simple string, an int or a boolean is parsed as YAML and converted to JSON. If the property is not valid YAML then the property is set to an empty string.

This means that normal steps can now have properties that are actually YAML: the environment variable will be set to the JSON equivalent.

The unit tests for `config.go` currently fail because this change means that an unsupported step property value such as "3.14" will be passed to the step as JSON rather than as an empty string as now, which is a change in behaviour. We need to decide what behaviour we require here.


nigeldeakin:
This didn't gain favour.

PR #461, opened 2018-07-27
Wrkr-592 - Deprovision RDD for builds aborted/interrupted

lake-of-dreams:
Changes to ensure that any RDD configured for a build is Deprovisioned for any interrupts/error conditions not otherwise handled or during box cleanup - primarily for situations when a user clicks "Abort" on UI ~~and a SIGTERM is sent to wercker cli pod (Related change in KP -  [wercker/kiddie-pool#432](https://github.com/wercker/kiddie-pool/pull/432))~~. Added calls to deprovision at
-  "Finisher" for the pipeline
-  Registered a "rdd-cleanup" handler to call deprovision for any aborts - registered at global level and therefor for any aborts during setup environment or at any step a abort will result in RDD getting cleaned up
- During box cleanup
- If there is any error in provisioning RDD itself or during its verification
- We keep track of any deprovision request using a flag in rdd - `rddDeprovRequested` - so that Deprovision calls RDD API server only once per build
- Have been tested both locally and on dev.wercker for all known possible scenarios with aborts in setup environment, in any other step, terminating a wercker cli pod from command line etc and seems to be working as expected
- This is a prerequisite before we move RDD feature to production so as to avoid any "Dangling" vms


rdd/rdd.go
```diff
@@ -149,21 +156,36 @@ func (rdd *RDD) Provision(ctx context.Context) (string, error) {
...
 		log.Warning(errMsg)
+		return
 	}
+	log.Infof("Finished deprovisioning RDD for runID: %s", rdd.runID)
 }
 
 // verify the RDD url by connecting to it and retrieving its version
-func verify(ctx context.Context, rddDetails *rddDetails) error {
+func (rdd *RDD) verify(ctx context.Context, rddURI string) error {
+	log.Infof("Verifying RDD at %s", rddURI)
```
hatchan: Maybe make this a Debug statement

rdd/rdd.go
```diff
@@ -149,21 +156,36 @@ func (rdd *RDD) Provision(ctx context.Context) (string, error) {
 
 //Deprovision - Deprovisions a RDD previously provisioned for a build
 func (rdd *RDD) Deprovision() {
+	if rdd.rddDeprovRequested {
+		log.Infof("RDD Deprovisioning already requested for runID: %s", rdd.runID)
```
hatchan: Maybe make this a warn

PR #465, opened 2018-07-31
add support for OCI object storage

mgianatagh:
https://jira.oraclecorp.com/jira/browse/WRKR-672

Add support for a new file store - OCI Object Storage

This change is required for managed runners


cmd/flags.go
```diff
@@ -170,6 +173,17 @@ var (
 		cli.StringFlag{Name: "aws-region", Value: "us-east-1", Usage: "AWS region to use for artifact storage."},
 	}
 
+	// OCI bits
+	OCIFlags = []cli.Flag{
+		cli.StringFlag{Name: core.OCI_TENANCY_OCID, Usage: "Unique identifier for OCI tenancy.", EnvVar: "wkr_tenancy_ocid"},
```
hatchan: Why are all the envvars prefixed with wkr and are not all upper case, which is the convention with envvars?

hatchan: Guess this has to do with the prefix config

hatchan: We should choose one over the other, as some of these flags are not used (other than logging). So providing a flag in the form of `--oci-tenancy-ocid=value` will not be picked up.

I would prefer everything to go through our own flags and use our conventions: 
- Values can come in from envvar and are overridden by cli flags.
- EnvVars start with `WERCKER_OCI_`.
- EnvVars are all uppercase.

hatchan: Maybe use this one: NewRawConfigurationProvider

mgianatagh: Ok, I will change the EnvVars to be uppercase and  start with WERCKER_OCI_  The pre-existing implementation of object store support in Grappler requires the lowercase "wkr_xxx" env vars.  Do you feel I should change grappler to use uppercase env vars?

hatchan: I think the third comment is more important to deal with first.

hatchan: Does that mean that you will only rely on our library for getting the values from the envvars?

mgianatagh: After looking at the grappler code my suggestion of also having grappler use uppercase env vars does not make sense.  Grappler is only providing a hard coded prefix of "wkr_" to the OCI api's, the names and case of all the expected environment variables for the OCI api are then based off the prefix.  Perhaps we should change our prefix to be "wercker_oci_" instead?

gilbode: Why should grappler be lower case?  My suggestion would be make this consistent across the board and just change what grappler does to be consistent - i.e. use upper case.

mgianatagh: For OCI the case of the variables are outside the control of grappler.  When grappler calls ocisdkcomm.ConfigurationProviderEnvironmentVariables(OciEnvVarPrefix, "") the resulting variables are all lowercase.

The grappler command line does not have any parameters to specify the environment variables.  It expects that they are set prior to issuing the grappler command.

This appears to be consistent with how S3 support works.  The AWS environment variables are expected to be set prior to issuing the grappler command.  And they happen to be required to be in uppercase.

So I don't think there is anything to change for grappler.  But we can have kiddie-pool expect environment variables of the form WERCKER_OCI_XXXX and internally KP will define a lowercase version of those variables before calling grappler.

hatchan: why do we use `ConfigurationProviderEnvironmentVariables` in the first place? Like I said in a previous comment, now we the possibility of two configurations.

mgianatagh: @desagar can you weigh in on these comments?  The object store code I am using was originally copied from the grappler object store implementation.  From my limited knowledge of these API's for OCI there does not appear to be any option to create an uppercase form of the environment variables.  In the file configuration.go of the oci-go-sdk the environment variables are hard coded to be lowercase.

desagar: I don't see any reason why we can't switch grappler to use a "raw configuration provider". We simply didn't notice its existence at the time, and it looks like a good idea. 

As Chen said, we can use either a composing config provider, or to completely have control over the values passed into OCI, we can choose to use the raw config provider, in which case we can name the env vars whatever we like.

mgianatagh: Thank you Deva & Chen!

PR #468, opened 2018-08-01
Docker run fetch

agnihotrigaurav:
Error in case of image pull is not handled in docker-run. It has caused below issues.
01) If docker run not able to pull images and image still exist locally, then docker-run steps does not fail.

Solution
01) For each run docker run tries to pull latest image from repository.
02) There is one exception to above, Docker run tries to check if there exist image with RunID+image name exist on machine, If same exist then docker-run does not call pull images(Image might be built in same pipeline, and user tries to run same for testing)
03) In case pull image fails, image name has been added in error message.

Staging testing
01) docker-build and docker run has same images. docker run would not call fetch method to pull image.
https://dev.wercker.com/gauagnih/wercker-docker-build/runs/build/5b619094a69a26000724c8a0
=================================================================== 

02) docker-build and docker run has different images. docker run would call fetch method to pull image
https://dev.wercker.com/gauagnih/wercker-docker-build/runs/build/5b6190d472c2aa0007697034
=================================================================== 

03) docker run has images that does not exist. Correct error message would be displayed.
https://dev.wercker.com/gauagnih/wercker-docker-build/runs/build/5b61912b72c2aa000769703b?step=5b61913e2e48810007e59564


docker/box.go
```diff
@@ -708,7 +708,7 @@ func (b *DockerBox) Fetch(ctx context.Context, env *util.Environment) (*docker.I
 	}
 	err = client.PullImage(options, authConfig)
 	if err != nil {
-		return nil, errors.Wrap(err, "fetch failed to pull image")
+		return nil, errors.Wrapf(err, "fetch failed to pull image %s", b.Name)
 	}
```
nigeldeakin: @jwells131313 has a PR https://github.com/wercker/wercker/pull/469 which adds a new function `getImageName` which returns `fmt.Sprintf("%s:%s", b.ShortName, b.tag)` instead of `b.Name`, and is used in an error message. Perhaps we should use that here?

docker/docker_run.go
```diff
@@ -182,11 +183,12 @@ func getCorrectImageName(env *util.Environment, s *DockerRunStep) (string, error
 	}
 
 	// local image should exists with a prepend of build id.
-	image, err := client.InspectImage(s.options.RunID + i)
```
nigeldeakin: Please update the comment to say something like "determine whether the specified image-name corresponds to an image created by a prior docker-build step (in which case image-name will be pre-pended by the build id). If it isn't assume it's a normal docker image"

docker/docker_run.go
```diff
@@ -215,13 +218,20 @@ func (s *DockerRunStep) Execute(ctx context.Context, sess *core.Session) (int, e
 		s.logger.Errorln("Error in creating a box from boxConfig ", boxConfig)
 		return 1, err
 	}
+
+	// In case docker-build is run before docker-run, Its possible that build image is not pushed to repository.
```
nigeldeakin: I suggest changing the comment to "Pull the image from the registry unless it was created by a prior docker-build step"

docker/box.go
nigeldeakin: On second thoughts I see that there is a line earlier which sets 
```
	b.Name = fmt.Sprintf("%s:%s", b.repository, b.tag)
```
so I withdraw my comment. `b.Name` is fine.

docker/docker_run.go
agnihotrigaurav: Done.

docker/docker_run.go
agnihotrigaurav: Done.

docker/box.go
jwells131313: truthfully for an error message I would prefer the shorter name (either b.ShortName or use my new function).  I doubt the repository name would be of interest and there is value in brevity here

PR #471, opened 2018-08-03
Experimental support for running workflows in yml locally

gsserge:
* Validation of the workflows will be implemented in a separate PR.
* Integration tests will be in another PR.
* For simplicity, output directories are copied between runs. We can think about symlinking for efficiency later.

A few example workflows to demonstrate the implemented functionality
```
workflows:
  - name: seq
    pipelines:
      - name: mybuild
        pipelineName: build
      - name: echoa
        requires:
          - mybuild
      - name: readone
        requires:
          - echoa

  - name: fanin
    pipelines:
      - name: build
      - name: echoa
        requires:
          - build
      - name: echob
        requires:
          - build
      - name: readfile
        requires:
          - echoa
          - echob

  - name: faninOne
    pipelines:
      - name: build
      - name: echoa
        requires:
          - build
      - name: echob
        requires:
          - build
      - name: readone
        artifactPipeline: echob
        requires:
          - echoa
          - echob
```


cmd/workflow.go
```diff
@@ -0,0 +1,202 @@
...
+	"github.com/wercker/wercker/util"
+	"golang.org/x/net/context"
+	"gopkg.in/mgo.v2/bson"
+)
+
+// cmdWorkflow is the main driver for running a workflow.
+func cmdWorkflow(ctx context.Context, opts *core.WorkflowOptions, dockerOptions *dockerlocal.Options) error {
+	// PipelineName->RunId map to keep track of which pipelines have ran
+	// and their runIDs.
+	executionStatus := map[string]string{}
```
flenter: Looking at the name I expected this to hold values like `not_started`, `running`, `finished`. How about something like `pipelineRunMap`?

PR #477, opened 2018-08-09
No hidden var interpolation

agnihotrigaurav:
Problem :- Password field can have special char (like $), which are interpolated on shell. OCIR passwords generally has $ sign because of which docker-push step is failing.

Solution :- Generally password are protected strings. In code they are evaluated separately as Hidden environment var.
While creating export statement and making (`export %s= '%s'`, key, e.Map[key]) change, special characters(like $) in Hidden list are not evaluated. 

Impact :- Users who were using interpolation feature on shell for protected variables would be impacted. I belive few users would be using interpolation feature in protected variables, so impact will be minimal.

Testcase executed are as follows 
a) When password contains $ (ChangeMe$1) and its a protected variable defined in UI

	01) password = ChangeMe$1 . Result=Pass
		https://dev.wercker.com/gauagnih/wercker-docker-push/runs/build/5b6abf5375e29b0006483331
	
	02) password = K-_&V4f>M.j^S`kf$Qf2+It1 . Result=Pass
		https://dev.wercker.com/gauagnih/wercker-docker-push/runs/build/5b6c095f75e29b000648378d

	03) password = K-_&V4f>M.jSkf$Qf2+It1 . Result=Pass
		https://dev.wercker.com/gauagnih/wercker-docker-push/runs/build/5b6c083978f7dd00070062eb

b) When password contains $ (ChangeMe$1) and its a public variable defined in UI. Testcase would fail.
    https://dev.wercker.com/gauagnih/wercker-docker-push/runs/build/5b6abf1a75e29b0006483322

### DIFF
--- a/core/pipeline.go
+++ b/core/pipeline.go
@@ -244,7 +244,7 @@ func (p *BasePipeline) ExportEnvironment(sessionCtx context.Context, sess *Sessi
 	// Export the hidden variables separately
 	sess.HideLogs()
 	defer sess.ShowLogs()
-	exit, _, err = sess.SendChecked(sessionCtx, p.Env().Hidden.Export()...)
+	exit, _, err = sess.SendChecked(sessionCtx, p.Env().Hidden.ExportHidden()...)
 	if err != nil {
 		return err
 	}
--- a/docker/box.go
+++ b/docker/box.go
@@ -389,7 +389,7 @@ func (b *DockerBox) RecoverInteractive(cwd string, pipeline core.Pipeline, step 
 
 	env := []string{}
 	env = append(env, pipeline.Env().Export()...)
-	env = append(env, pipeline.Env().Hidden.Export()...)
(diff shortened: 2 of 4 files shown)


agnihotrigaurav:
Above solution seems will minimize the impact. @gilbode suggested we should not interpolate web Environment variables. While looking into code i found that hidden variables are stored separately. So this change only stops that interpolation. 

@gilbode Shall we go ahead with above solution or we changed it to the one mentioned in comments?.

gilbode:
@agnihotrigaurav  Please do as i asked and apply to all variables passed from the envvars service (hidden or not) - i just had an example where i had a password in an environment variables with special characters in it and the pipeline hung in setup environment.  Any environment can have special characters in it.

agnihotrigaurav:
Closing as another PR raised which stops interpolation of all WEB variables

PR #478, opened 2018-08-09
Add InternalBuildFlags to workflow command.

gilbode:
Enables flags such as direct-mount, expose-ports, attach-on-error, enable-volumes, enable-dev-steps.


nigeldeakin:
I've tried this and the `--attach-on-error` parameter now works.

PR #479, opened 2018-08-10
Wrkr 828 - Retry CreateContainer to avoid no such image error

lake-of-dreams:
Added new function that invokes docker `CreateContainer` with a timeout based retry and replaced the call to `CreateContainer` for `box`, `service` and `docker_run` (These are all the places from where we `CreateContainer` at the moment from wercke cli) - Tested use case of box,service and docker_run with and without RDD, added a unit test case that simulates `no such image`


docker/docker_client_fsouza.go
```diff
@@ -225,3 +232,27 @@ func (c *DockerClient) ExecOne(containerID string, cmd []string, output io.Write
 
 	return nil
 }
+
+// CreateContainerWithRetries create a container - retry on "no such image" error
+func (c *DockerClient) CreateContainerWithRetries(opts docker.CreateContainerOptions) (*docker.Container, error) {
+	timeout := time.After(DefaultTimeoutForCreateContainer)
+	tick := time.Tick(500 * time.Millisecond)
+	for {
```
hatchan: Can't we just try `x` times before giving up. Seems way simpler than this channel approach.

lake-of-dreams: In this situation timeout based retry is more appropriate with a brief interval between retries and in my opinion much cleaner than having tried random x no. of times. The root cause of this issue is still speculative - https://github.com/moby/moby/issues/30658 and most likely due to image not present in docker cache - and therefore we should wait for sometime before retrying again in an anticipation of getting any caching issues resolved. If we just retry x no. of times, without waiting between retries, I do not see any benefit as we are not allowing the intermittent problem, if any, to get resolved. If you meant we should retry x no. of times, with `sleep` for some time between retries, than this approach is much better than that.

lake-of-dreams: In this situation timeout based retry is more appropriate with a brief interval between retries and in my opinion much cleaner than having tried random x no. of times. The root cause of this issue is still speculative - https://github.com/moby/moby/issues/30658 and most likely due to image not present in docker cache - and therefore we should wait for sometime before retrying again in an anticipation of getting any caching issues resolved. If we just retry x no. of times, without waiting between retries, I do not see any benefit as we are not allowing the intermittent problem, if any, to get resolved. If you meant we should retry x no. of times, with `sleep` for some time between retries, than this approach is much better than that.

hatchan: A sleep between retries is kinda implied, yes. We would probably want to exponentially increase the sleeps, so that we're not hammering the API either.

lake-of-dreams: yeah than that is what this code is doing, retrying the API every 1/2 an second, only thing is instead of sleep it is doing a `tick` and we have an option of timeout rather than no. of retries. do you mean we should increase the time of tick from 500ms? to 1 s?

hatchan: if the CreateContainer is timing out for whatever reason, then this approach won't work either.

hatchan: Also the retry might be 0,5 seconds apart but it is not guaranteed. As the timer just keeps ticking every 0,5 seconds, not 0.5 seconds after this execution.

hatchan: Also `time.Tick` leaks as there is no way of stopping it.

hatchan: Finally don't call this a circuit breaker as that is something else

lake-of-dreams: yes I did not mean circuit breaker in the traditional sense here, and even if retry is not guaranteed at 0.5 seconds , then also it is fine as long as a retry is already in progress when it ticks. Only two valid points against this approach hat I can see from this discussion is that leak in `time.Tick` and if `CreateContainer` is taking longer , it does not timeout which is what I was expecting. I will check both of these

hatchan: If `CreateContainer` is timing out as expected, then both solutions take the same amount of time. Another big valid point is that this approach is that a simple loop is way easier to understand.

lake-of-dreams: I checked and in order to be able to `timeout` on a long running `CreateContainer` , without the need of a leaky `time.Tick` - `CreateContainer` will need to be called from a `goroutine` with channels that emit `container` and `error` and in main for, select - we check for `timeout`, and channels for `container` and `error`

```
        errorChan := make(chan error, 1)
	containerChan := make(chan *docker.Container, 1)

	go func() {
		for {
			container, err := c.CreateContainer(opts)
			if err != nil {
				switch err {
				case docker.ErrNoSuchImage:
					time.Sleep(500 * time.Millisecond)
					continue
				default:
					errorChan <- err
					return
				}
			}
			containerChan <- container
		}
	}()

	timeout := time.After(DefaultTimeoutForCreateContainer)

	for {
		select {
		case <-timeout:
			return nil, errors.New("Timed out trying to create container")
		case container := <-containerChan:
			return container, nil
		case err := <-errorChan:
			return nil, err
		default:
			// NooP

		}
	}
```

Now it may not be too useful to check a `timeout` for the cases where this change is made, because in all the scenarios it is highly unlikely that `CreateContainer` will ever timeout, 
 - We verify docker daemon connectivity for both RDD and non-RDD cases before the logic reaches this point
- We already either pull image from which container is being created or it is already built using `internal/docker-build` step 
so it may not be that useful to introduce this logic to just check a `timeout` scenario which may not occur at all and therefore a retry loop will be more appropriate.

Understand your opinion that a simple loop is much easier to understand but the fact is that if indeed there is a case of long running API call, a channel based approach described above might actually be helpful

PR #481, opened 2018-08-14
Wrkr 807 - Display message while pushing images built using internal/docker-push

lake-of-dreams:
1/ Display a message while pushing images built using `internal/docker-build` step
2/ Some refactoring regarding use of `image-name` property 
3/ Also make argument of `tagAndPush` as `imageRef` and not `imageID` because in case of images built by `internal/docker-build` we are passing in image name while in normal scenarios it is imageID
4/ Tested locally and using canary


docker/docker.go
```diff
@@ -803,10 +806,16 @@ func (s *DockerPushStep) Execute(ctx context.Context, sess *core.Session) (int,
 		if err != nil {
 			return -1, err
 		}
-		imageID = idResponse.ID
-		s.logger.WithField("Image", imageID).Debug("Commit completed")
+		imageRef = idResponse.ID
+		s.logger.WithField("Image", imageRef).Debug("Commit completed")
```
hatchan: field keys need to be camelcase

PR #499, opened 2018-08-28
Escape special characters in bash exports

gilbode:
Including backtick causes setup environment to hang.  Also escape other characters whose special meaning is preserved in bash double quotes (with the exception of $ - treat interpolation separately).


agnihotrigaurav:
@gilbode I think it would be redundant once interpolation of UI variables are stoped.

gilbode:
@hatchan - if my test is to be believed, the case you mentioned did not break.  @agnihotrigaurav - this is slightly different than interpolation - it's not redundant exactly (though depending on how you implement could have the same effect).

PR #503, opened 2018-08-30
Add a new pipeline to run the integration tests in Wercker

nigeldeakin:
This PR defines a new pipeline `integration-tests` which runs the integration tests (by which I mean the script `run-tests.sh`).  

There are four integration tests which do not currently work when run in a Wercker pipeline are automatically skipped: to run these then the user needs to run `test-all.sh` locally. 

This new pipeline is designed to run after the existing `build` pipeline. Since the new pipeline needs access to the wercker tests I have added an extra step to the existing `build` pipeline to copy the source code to the output directory. 

After this change is released I plan to modify the workflow in staging and prod to run the new pipeline automatically. Note however that such a change can't be done as part of this PR. The new pipeline takes about 20 minutes to run 40 tests.  (In comparison, `test-all.sh` takes 20 minutes on my Macbook Pro to run 44 tests)

Although it is a long-term goal, at this stage I do not propose that we make make PRs conditional on the new pipeline passing. That's something to consider at a later stage.

You can see the new pipeline being run at https://app.wercker.com/nigeldeakin/wercker/runs

SInce this does not change anything that we actually release I have not recorded this change in the changelog.


nigeldeakin:
Technical details:

Before running these tests the `integration-tests` pipeline starts a docker daemon running in the pipeline container and sets `DOCKER_HOST` accordingly. This means the pipeline has to have `docker: true` specified because that causes the pipeline container to run in privileged mode, and a docker container is not allowed to start a docker daemon unless the container is started in privileged mode.

Note that the pipeline can't simply set `docker:true` (to enable direct docker access) and use the default setting for `DOCKER_HOST`, which is to point to the daemon in which the pipeline container is running. This is because these tests use the Wercker CLI, which will start containers in the daemon specified by DOCKER_HOST, and Wercker CLI (when run from the command line rather than triggered by KP) requires that any such containers are able access the same file system that the Wercker CLI uses.

I've pasted the

nigeldeakin:
If anyone knows how to avoid the `Unable to locate plugin: , retrying in 2s` messages, let me know.

test-all.sh
```diff
@@ -96,8 +109,19 @@ testScratchPush () {
 
 
 runTests() {
-  source $testsDir/rdd/test.sh || return 1
-  source $testsDir/rdd-volumes/test.sh || return 1
+
+  #  The following tests must be skipped when run in a wercker pipeline 
+  if [ -z ${WERCKER_ROOT} ]; then 
+    # The rdd tests cannot be run in wercker because the pipeline cannot connect to the daemon
+    source $testsDir/rdd/test.sh || return 1
```
lake-of-dreams: So when you run in box `docker:dind` - it does not have `/var/run/docker.sock` as DOCKER_HOST - because I think when `dcoker:true` and since there will be no `rdd-service-uri` specified - the rdd tests should have started to use local docker daemon  - which is docker daemon of `docker:dind`? What happens when in your `integration-test` pipeline you start `dockerd` without overriding `--host`? Are you able to run rdd tests then?

test-all.sh
```diff
@@ -96,8 +109,19 @@ testScratchPush () {
...
+    source $testsDir/rdd-volumes/test.sh || return 1
+    # local services test cannot be run in wercker because the pipeline cannot connect to the local service (despite being in the same network)
+    basicTest "local services"    build "$testsDir/local-service/service-consumer" --docker-local || return 1
```
lake-of-dreams: Has this been investigated why we are not able to connect to service or is there a separate Jira for that?

wercker.yml
```diff
@@ -83,6 +84,57 @@ build:
...
+        name: Copy and test the wercker executable
+        code: |
+            export GOPATH=$WERCKER_SOURCE_DIR/go
+            cp -p $WERCKER_SOURCE_DIR/artifacts/latest/linux_amd64/sentcli $GOPATH/src/github.com/wercker/wercker/wercker
+            $GOPATH/src/github.com/wercker/wercker/wercker --version            
+
+    - script:
+        name: Start the docker daemon
+        code: |
+            dockerd --host=tcp://0.0.0.0:2375 --storage-driver=overlay &
```
lake-of-dreams: Why do we have to change `--host` here?

test-all.sh
nigeldeakin: In the `integration-tests` pipeline, the "integration-tests" step starts the daemon and sets `DOCKER_HOST` to `tcp://127.0.0.1:2375`. That allows subsequent calls to `wercker` and `docker version` etc to connect to the new daemon. (It doesn't use `/var/run/docker.sock` because that's already in use for the outer daemon)

When a test invokes `wercker` to run a pipeline, then `DOCKER_HOST` within the new (nested) pipeline has the same value of `DOCKER_HOST`. That's normal behaviour when running the `wercker` command from the command line. 

However that value of `DOCKER_HOST` doesn't actually work within the new (nested) pipeline. I'm not sure how to resolve that. Suggestions welcome; I plan to log this as an issue to be resolved later.

test-all.sh
nigeldeakin: Yes, I have spent some time investigating it but been unable to fix it. I plan to log this as an issue to be resolved separately. 

A local service is essentially one which is defined in a separate `wercker.yml`. Before a pipeline that uses a local service is started, wercker is run in a kind-of-recursive manner to  execute the service pipeline and save it as an image. The saved image is then started just like any other service before the main pipeline is started.  However when this is all run in the `integration-tests` pipeline I see that the main pipeline (the one that uses the local service) cannot connect to the service. I haven't discovered why: normal services work just fine, and once the temporary image is saved the local service appears to be started using the same code as normal services. I have confirmed that both containers are using the same docker network, and are running in the expected docker daemon.

I have deferred further investigation on this; I will log a JIRA to cover this.

wercker.yml
nigeldeakin: If we don't change it then the nested daemon will try to connect to `/var/run/docker.sock` which will fail because it is already being used by the outer daemon.

test-all.sh
lake-of-dreams: It doesn't use `/var/run/docker.sock` because that's already in use for the outer daemon >> by outer daemon you mean RDD where this pipeline container is running? does it mean that when you run container using `docker:dind` image and with `--privileged` - `/var/run/docker.sock` inside container is same as `/var/run/docker.sock` of the host?

nigeldeakin: I'll run a test to demonstrate what happens if the nested daemon is started with the default host settings. (Happy to discuss and explore this area btw)

wercker.yml
lake-of-dreams: seems odd - I just did
```
 docker run --privileged -ti docker:dind /bin/sh
 dockerd &
```
and it displayed the message
```
INFO[2018-08-31T08:42:54.556123313Z] API listen on /var/run/docker.sock 
```

After which I did

```
docker ps -a
```

And it did not show any containers from host

lake-of-dreams: If you do 
```
- script:
        name: Start the docker daemon
        code: |
            dockerd &
            export DOCKER_HOST=unix:////var/run/docker.sock
            docker ps -a
```

i.e. override the DOCKER_HOST set by RDD logic, then **may be** you should be able to run RDD tests

nigeldeakin: I've remembered now. The problem this is trying to avoid is when you run `wercker` locally, when wercker's existing logic typically decides that the daemon is available using `unix:///var/run/docker.sock`. 

In this case, when a pipeline is started with `docker: true`, the daemon socket is mounted when the pipeline container is started, and  `DOCKER_HOST` in the pipeline is set to `unix:///var/run/docker.sock`.

If this pipeline then attempts to start `dockerd` it will fail because `/var/run/docker.sock` is in use. This is demonstrated in the following:

```
docker run --privileged -v /var/run/docker.sock:/var/run/docker.sock -ti docker:dind /bin/sh
/ # dockerd
WARN[2018-08-31T09:18:03.081217400Z] could not change group /var/run/docker.sock to docker: group docker not found
Failed to load listeners: can't create unix socket /var/run/docker.sock: device or resource busy
```
We might possibly be able to avoid this by changing the logic that wercker uses to work out how to connect to the daemon. Currently it tries first to connect using the socket - which usually works when running wercker on a laptop. However we could change this to try a tcp connection first. If that worked we wouldn't need to mount the socket on the pipeline container, which would allow the nested daemon to use it.

nigeldeakin: Alternatively we could leave wercker unchanged and simply expect people who want to run this pipeline on their laptop to set `DOCKER_HOST` to use a tcp connection. (If the local daemon doesn't expose a port - as is the case with Docker for Mac - they can run a proxy). We could add a test at the start of the pipeline and fail if the user has not done this.

That looks a good way forward. Updates coming soon.

test-all.sh
nigeldeakin: (Discussion continues in [the other thread](https://github.com/wercker/wercker/pull/503#pullrequestreview-151291754))

nigeldeakin:
Prompted by the helpful comments from @lake-of-dreams I have made some more changes. The "nested" daemon now listens on a socket rather than on a tcp port, which allows nested containers running in that daemon to connect to it (when they couldn't do before).

This change makes the two rdd tests work, so they no longer need to be skipped. 

To avoid any clash with the socket used by the "outer" daemon, the "nested" daemon uses a different socket name: /var/run/docker2.sock

test-all.sh
```diff
@@ -8,7 +8,20 @@
 # that is not already in the docker daemon these tests will fail with "image not found"
 # The function pullImages below pulls a list of specified images before running the tests. Update it if needed.
 #
-# To run the tests
+# This script can be run either from the command line or in a Wercker pipeline.
+# Thise tests that cannot be run in a Wercker pipeline are skipped automatically.
```
hatchan: `Thise` -> `The`

nigeldeakin: Now fixed (to `Those`)

PR #506, opened 2018-09-04
Changes to make envvars test pass

nigeldeakin:
The tests in `tests/projects/envvars` fail (when run locally using `test-all.sh` as well as in a wercker pipeline). The following changes make the test pass.

Only approve this PR if we expect `!` characters to be converted to `\!` (which seems unlikely). If not then there is a bug in `wercker/wercker`.

`test-all.sh` sets the following:
```
  export X_BACKSLASH='back\slash'
  export X_BACKTICK='back`tick'
  export X_BANG='ban!g'
  export X_BANG_ESCAPED='ban\!g'
```

* In `wercker.yml`, I have changed `ESCAPED_BANG` to `BANG_ESCAPED` since that's the name of the variable set in `test-all.sh`

* `X_BANG='ban!g'` gets converted to  `ban\!g` rather than `ban!g` so I have changed `wercker.yml` to make the test pass. I suspect this is not what is expected.

* `X_BANG_ESCAPED='ban\!g'` gets converted to  `ban\\!g` rather than `ban!g` so I have changed `wercker.yml` to make the test pass. I suspect this is not what is expected.


gilbode:
I must have done something wrong bc these passed for me before I merged.

hatchan:
I would expect `ban\!g` to get changed to `ban\\\!g` though.

gilbode:
The last one is what’s wrong - the double escape.  I will not be able to fix this for a few hours.  You can either revert my merge or remove all the escaping of ! In the actual code (should be one line) and modify the tests accordingly.

gilbode:
See #508

nigeldeakin:
I'll close this as it has served its purpose. Now superseded by https://github.com/wercker/wercker/pull/508.

PR #513, opened 2018-09-10
Wrkr 989 - Validate tag format before push

lake-of-dreams:
An issue was encountered where if any of the tags described in `tag` property of `internal/docker-push` step is on invalid format, docker push will fail bit no error will be displayed in screen. This change to validate tag format before performing a docker push and return any validation errors so that these will be displayed on screen


docker/docker.go
```diff
@@ -776,6 +777,10 @@ func (s *DockerPushStep) Execute(ctx context.Context, sess *core.Session) (int,
 	s.logger.Debugln("Init env:", s.data)
 
 	var imageRef = ""
+	repositoryNameWithTag, err := validateAndGetRepositoryNameWithTag(s.repository, s.tags)
+	if err != nil {
```
nigeldeakin: This change validates the first tag, which is used when committing the container.

However in `tagAndPush` the code loops round each of the specified tags. For each such tag, the image is tagged with this tag and then pushed. Do we need to validate those tag value as well?

docker/docker.go
```diff
@@ -917,3 +919,24 @@ func (s *DockerPushStep) ShouldSyncEnv() bool {
 	}
 	return true
 }
+
+// validateAndGetRepositoryNameWithTag - Verifies that `repository` and all of the `tags` are in valid format
+// and returns an error if not. When all the tags are valid, returns `repository:tag` where `tag` is first
+// tag in the `tags` array
+func validateAndGetRepositoryNameWithTag(repository string, tags []string) (string, error) {
```
nigeldeakin: I found this a bit confusing: it appears to do two distinct things. 
* It validates all the tags, which are used later on in `tagAndPush`
* It constructs and returns a repository:tag string using the first tag for use when committing the container 
The name of the function reflects this dual purpose.

I wonder whether it would be clearer if these two things were separated into separate functions.

lake-of-dreams: Fixed, now `validateTags` only validate tags, we perform repository validation already in `InferRegistryAndRepository` - here we parse that because we need to pass the parse `reference.Named` instance corresponding to repository in  `reference.WithTag` - here we just double check  the error and don't need a separate function for this

lake-of-dreams:
Validates tag format using `reference.WithTag` - see https://github.com/docker/distribution/blob/642075f42c027b8a05b7d06f4f132939eb9ed902/reference/reference.go#L271 for all the validations it performs. Implemented for `docker-scratch-push` as well and test cases. Moved validation on top in `Execute` call for both the steps

docker/docker.go
```diff
@@ -101,12 +102,26 @@ func NewDockerScratchPushStep(stepConfig *core.StepConfig, options *core.Pipelin
 
 // Execute the scratch-n-push
 func (s *DockerScratchPushStep) Execute(ctx context.Context, sess *core.Session) (int, error) {
+
```
hatchan: Remove this empty line

docker/docker.go
```diff
@@ -917,3 +936,15 @@ func (s *DockerPushStep) ShouldSyncEnv() bool {
 	}
 	return true
 }
+
+// validateTags - Verifies that all of the `tags` are in valid format
```
hatchan: Please write this as a sentence: `validateTags verifies that...`

docker/docker.go
```diff
@@ -750,6 +762,22 @@ func (s *DockerPushStep) Fetch() (string, error) {
 // and pushes it to the configured registry when image-name property is not specified, in which case,
 // it tags and pushes an existing image built by a previous internal/docker-build step
 func (s *DockerPushStep) Execute(ctx context.Context, sess *core.Session) (int, error) {
+
```
hatchan: Remove this empty line

docker/docker.go
lake-of-dreams: Fixed

docker/docker.go
lake-of-dreams: Fixed

docker/docker.go
lake-of-dreams: Fixed

PR #515, opened 2018-09-13
Show error when publishing a step fails

gsserge:


nigeldeakin:
I have tested this and in this particular case the step now gives this message:
```
Unable to publish step to the registry: Unable to start publish flow: Received error: 409: {"error":"step already exists","message":"step already exists","code":6,"details":[]}
```
LGTM

PR #519, opened 2018-09-19
Correct  WERCKER_BUILD_URL and WERCKER_DEPLOY_URL

nigeldeakin:
This PR fixes https://jira.oraclecorp.com/jira/browse/WRKR-1048 (see the issue for more information on the change needed)

So instead of `WERCKER_DEPLOY_URL` being set to
```
https://app.wercker.com/#nigeldeakin/testcase1/deploy/5b9fab2bc81a8a00066d8c1b
```
it is now set to
```
https://app.wercker.com/nigeldeakin/testcase1/runs/deploy/5b9fab2bc81a8a00066d8c1b
``` 
and instead of `WERCKER_BUILD_URL` being set to
```
https://app.wercker.com/#nigeldeakin/wercker-tests/build/5b9fd712c81a8a00066da3ff
```
it is now set to
```
https://app.wercker.com/nigeldeakin/wercker-tests/runs/build/5b9fd712c81a8a00066da3ff?step=5b9fd72482df1e000710c4ec
```

During tested I discovered that the URL always starts with `https://app.wercker.com`, even on staging. This value can be overridden by passing the (apparently deprecated) `base-url` parameter to wercker. However any fix to that would be a change to KP, not `wercker/wercker`.


hatchan:
That new `WERCKER_BUILD_URL` seems to not be right in the description. But I'm assuming that is just the description.

nigeldeakin:
This fix also fixes `WERCKER_RUN_URL`. I've re-tested.

PR #522, opened 2018-10-03
Supress build logs

agnihotrigaurav:
JiraID WRKR-1087 - suppress build logs.

Flag has been added, if flag is true then logs are not sent to standard output.


agnihotrigaurav:
Currently run logs do not go to kibana because of security issue. If we suppress user logs from going to standard output, we can reenable fluentd to send standard output/error messages to Kibana. 
In case network error exist, we can raise alert message so that corrective action can be taken. 

User logs are captured by LiteralLogHandler emmiter, In case flag is true, we do not instantiate this loghandler.

markxnelson:
what impact would this have on the user though? how do we get the logs into the web ui? does that rely on them being written to stdout?

cmd/flags.go
```diff
@@ -128,6 +128,7 @@ var (
 	InternalDevFlags = []cli.Flag{
 		cli.BoolTFlag{Name: "direct-mount", Usage: "Mount our binds read-write to the pipeline path."},
 		cli.BoolFlag{Name: "expose-ports", Usage: "Enable ports from wercker.yml beeing exposed to the host system."},
+		cli.BoolFlag{Name: "suppress-build-logs", Usage: "If specified, output from steps is not logged", Hidden: false},
```
hatchan: Make these `Hidden: true`

hatchan: (also applies to the other ones)

agnihotrigaurav:
There is another bug WRKR-1088 which involves loading data to web UI. Recently logs were disabled from going to web UI because of security concern. As discussed with @nigeldeakin and @hatchan to reenable logging we first need to disable user logs.

cmd/runner.go
```diff
@@ -108,12 +108,14 @@ func NewRunner(ctx context.Context, options *core.PipelineOptions, dockerOptions
...
 	}
-
-	l, err := event.NewLiteralLogHandler(options)
-	if err != nil {
-		logger.WithField("Error", err).Panic("Unable to event.LiteralLogHandler")
+	var l *event.LiteralLogHandler
+	if !options.SuppressBuildLogs {
+		l, err = event.NewLiteralLogHandler(options)
+		if err != nil {
+			logger.WithField("Error", err).Panic("Unable to event.LiteralLogHandler")
```
hatchan: While you're changing things here: can you change `WithField("Error", err)` to `WithError(err)`

cmd/flags.go
```diff
@@ -128,6 +128,7 @@ var (
 	InternalDevFlags = []cli.Flag{
 		cli.BoolTFlag{Name: "direct-mount", Usage: "Mount our binds read-write to the pipeline path."},
 		cli.BoolFlag{Name: "expose-ports", Usage: "Enable ports from wercker.yml beeing exposed to the host system."},
+		cli.BoolFlag{Name: "suppress-build-logs", Usage: "If specified, output from steps is not logged", Hidden: true},
```
nigeldeakin: Should this be "...output from steps is not sent to stdout" ? 

If the output from steps is written using the logging API then use the word "logged". if it isn't then don't.

cmd/flags.go
```diff
@@ -143,6 +144,8 @@ var (
 	InternalBuildFlags = []cli.Flag{
 		cli.BoolFlag{Name: "direct-mount", Usage: "Mount our binds read-write to the pipeline path."},
 		cli.BoolFlag{Name: "expose-ports", Usage: "Enable ports from wercker.yml beeing exposed to the host system."},
+		cli.BoolFlag{Name: "suppress-build-logs", Usage: "If specified, output from steps is not logged", Hidden: true},
```
nigeldeakin: Should this be "...output from steps is not sent to stdout" ?

cmd/flags.go
```diff
@@ -158,6 +161,7 @@ var (
 	// Flags for advanced deploy settings
 	InternalDeployFlags = []cli.Flag{
 		cli.BoolFlag{Name: "expose-ports", Usage: "Enable ports from wercker.yml beeing exposed to the host system."},
+		cli.BoolFlag{Name: "suppress-build-logs", Usage: "If specified, output from steps is not logged", Hidden: true},
 		// deprecated
```
nigeldeakin: Should this be "...output from steps is not sent to stdout" ?

cmd/runner.go
```diff
@@ -108,18 +108,20 @@ func NewRunner(ctx context.Context, options *core.PipelineOptions, dockerOptions
...
 	}
-
-	l, err := event.NewLiteralLogHandler(options)
-	if err != nil {
-		logger.WithField("Error", err).Panic("Unable to event.LiteralLogHandler")
+	var l *event.LiteralLogHandler
+	if !options.SuppressBuildLogs {
+		l, err = event.NewLiteralLogHandler(options)
+		if err != nil {
+			logger.WithError(err).Panic("Unable to event.LiteralLogHandler")
```
nigeldeakin: The message doesn't make sense. Is there a word missing?

cmd/flags.go
agnihotrigaurav: I suppose Logged is more generic term. It depends on Emitter to send logging output to stdout or some other place. If any changes done to logging output then statement would be inconsistent.

cmd/runner.go
agnihotrigaurav: Seems like old issue. changed below logger to 
logger.WithError(err).Panic("Unable to create event.LiteralLogHandler")
logger.WithError(err).Panic("Unable to create event.ReportHandler")

nigeldeakin:
OK. Just waiting for a report on testing.

agnihotrigaurav:
Updated Jira-1087 with test case that was run for verification of the bug. 
Integration test is failing because of elasticsearch image issue.

nigeldeakin:
I note that you have written some words about testing at https://jira.oraclecorp.com/jira/browse/WRKR-1087?focusedCommentId=35125260&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-35125260

Are all these tests manual or are any automated?

It's not clear what code changes those tests relate to, but since you mention kibana I presume that is a complete test with wercker logs being sent to kibana. Therefore can you please confirm:
1. Step output is still seen in the run screen in the web UI?
2. Step output is not written to stdout when wercker CLI is run locally (without the flag)
3. Step output is written to stdout when wercker CLI is run locally (with the flag)
4. Wercker logging is visible in Kibana
5. Step putput is not visible in Kibana

nigeldeakin:
@agnihotrigaurav Thanks. I just wanted to be verify that had considered all these cases. LGTM.

PR #523, opened 2018-10-04
wrkr-980 : Creating WERCKER_GIT_TAG env var

snanurag:
Task : https://jira.oraclecorp.com/jira/browse/WRKR-980

Please review all these 3 prs together.
- https://github.com/wercker/trigger/pull/153
- https://github.com/wercker/kiddie-pool/pull/454
- https://github.com/wercker/wercker/pull/523

### DIFF
--- a/core/options.go
+++ b/core/options.go
@@ -227,7 +227,12 @@ func guessGitBranch(c util.Settings, e *util.Environment) string {
 	if err != nil {
 		return ""
 	}
-	return strings.Trim(out.String(), "\n")
+	branch = strings.Trim(out.String(), "\n")
+	// In case of tag, branch output is HEAD.
+	if branch == "HEAD" {
+		return ""
+	}
+	return branch
 }
 
 func guessGitTag(c util.Settings, e *util.Environment) string {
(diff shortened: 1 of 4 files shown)


core/options.go
```diff
@@ -226,6 +227,59 @@ func guessGitBranch(c util.Settings, e *util.Environment) string {
...
+	os.Chdir(projectPath)
+
+	git, err := exec.LookPath("git")
+	if err != nil {
+		return ""
+	}
+
+	var out bytes.Buffer
+	cmd := exec.Command(git, "tag", "--points-at", "HEAD")
+	pipeCmd := exec.Command("tail", "-n", "1")
```
gsserge: Shouldn't `exec.LookPath("tail")` be used to locate the tail bin similar to git?

gsserge: Actually, I think there's no real need for using `tail -n 1` here. Just take the output from `git tag`, convert it to string and then use `strings.Split(outStr, "\n")[0]`.

snanurag: I have implemented and tested the changes. This build is failing because docker is not able to pull elasticsearch image. Gaurav and Benno are looking into that issue. I will merge it in master only after their fix is implemented.

snanurag:
Waiting on merge of https://github.com/wercker/wercker/pull/524

PR #531, opened 2018-10-30
Wrkr 1286 - Fix SyncEnvironment for alpine images

lake-of-dreams:
Earlier `env --null` is sent to container and then pipeline environment variables were being synced to the variable form the container taken from output of this command. However `alpine` based images does not support `--null` option and therefore we put a retry to send `env` - if `env --null` has failed.

Also there was a `set -e` that was sent with the `env --null` command which suppressed the non zero exit status code. That has been removed

Added a test case that invokes `internal/docker-push` which invokes `SyncEnvironment` and derives value of `REPO` from the environment variables provided in a `ENVIRONMENT`. The test checks whether the step does not result in an error ( The step produces an error if `repository` is not provided or if is not in correct format.

Have been tested locally and as well using canary.

PS: This does not fix the issue when environment variables inside `box` based on alpine images contain a newline in the value. A separate jira is logged to fix that if required - https://jira.oraclecorp.com/jira/browse/WRKR-1336


tests/projects/sync-env-alpine/test.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+
+# This is intended to be called from wercker/test-all.sh, which sets the required environment variables
+# if you run this file directly, you need to set $wercker, $workingDir and $testDir
```
nigeldeakin: Don't forget to add this test to `wercker/test-all.sh`

lake-of-dreams: done - thanks for pointing out

core/pipeline.go
```diff
@@ -267,32 +267,45 @@ func (p *BasePipeline) LogEnvironment() {
...
 	}
 
 	if exit != 0 {
 		return fmt.Errorf("Unable to sync environment, exit code: %d", exit)
 	}
 
-	// Concat every output line into a single string, then split on the null byte
-	full := strings.Join(output, "")
-	lines := strings.Split(full, "\x00")
+	if err != nil {
```
hatchan: It doesn't sound like we want to move error checking all the way over here. The error doesn't have to be related to the execution within the container, could also be connecting to the container.

tests/projects/sync-env-alpine/wercker.yml
```diff
@@ -0,0 +1,8 @@
+build:
+  box: alpine
```
hatchan: We probably want to pin this on a specific alpine image, where we know this "problem" exists. Just in case they switch the `env` binary.

core/pipeline.go
hatchan: Or will this contain a object if `exitCode != nil`?

core/pipeline.go
```diff
@@ -267,32 +267,45 @@ func (p *BasePipeline) LogEnvironment() {
...
 	sess.HideLogs()
 	defer sess.ShowLogs()
 
 	// 'env' with --null parameter, which prevents issues from overlapping \n
 	// inside the values.
-	exit, output, err := sess.SendChecked(sessionCtx, "set +e", "env --null", "set -e")
-	if err != nil {
-		return err
+	exit, output, err := sess.SendChecked(sessionCtx, "set +e", "env --null")
+	if exit == 0 && err == nil && output != nil && len(output) > 0 {
```
hatchan: The if statements seem pretty complex, we should try to remove the error check in here (see next comment) and try to capture some of these checks in variables, such as `hasOutput := output != nil && len(output) > 0`

core/pipeline.go
```diff
@@ -267,32 +267,45 @@ func (p *BasePipeline) LogEnvironment() {
...
 
 	sess.HideLogs()
 	defer sess.ShowLogs()
 
 	// 'env' with --null parameter, which prevents issues from overlapping \n
 	// inside the values.
-	exit, output, err := sess.SendChecked(sessionCtx, "set +e", "env --null", "set -e")
-	if err != nil {
-		return err
+	exit, output, err := sess.SendChecked(sessionCtx, "set +e", "env --null")
```
hatchan: We should still send ` "set -e"` to ensure that it is running in the same state as we left it. 

Regarding the comment in the in the PR description; we should not remove `"set +e"` as doing so would fail the shell session, and thus the wercker cli cannot send any more commands.

core/pipeline.go
```diff
@@ -267,32 +267,45 @@ func (p *BasePipeline) LogEnvironment() {
...
+	if exit == 0 && err == nil && output != nil && len(output) > 0 {
+		// Concat every output line into a single string, then split on the null byte
+		full := strings.Join(output, "")
+		lines = strings.Split(full, "\x00")
+	}
+
+	if exit != 0 {
+		if output != nil && len(output) > 0 && strings.HasPrefix(output[0], "env: unrecognized option: null") {
+			// send just 'env' if 'env --null' has failed (e.g. for alpine images)
+			exit, output, err = sess.SendChecked(sessionCtx, "set +e", "env")
```
hatchan: See previous comment, needs `set -e`

nigeldeakin:
I have confirmed that this change fixes https://jira.oraclecorp.com/jira/browse/WRKR-1050 (Internal steps in alpine prevent after-steps from running). That's because it stops `SyncEnvironment1 crashing.

core/pipeline.go
lake-of-dreams: For alpine, when we execute `exit, output, err := sess.SendChecked(sessionCtx, "set +e", "env --null", "set -e")`

We get

```
exit:
0

output:
[0]:"env: unrecognized option: null\n"
[1]:"BusyBox v1.28.4 (2018-07-17 15:21:40 UTC)

err:
nil
```

So `err` is `nil` and `exit` code is `0` - so only way we can know that the command has actually failed is by parsing the output

However when we execute `exit, output, err := sess.SendChecked(sessionCtx, "set +e", "env --null") - we get

```
exit:
1

output:
[0]:"env: unrecognized option: null\n"
[1]:"BusyBox v1.28.4 (2018-07-17 15:21:40 UTC)

err:
s:"Command exited with exit code: 1”

```

which is a more clearer indication that the command has actually failed.

However I understand your point as well that we do not want any side effects of running this - therefore will make changes to deduce that the command has actually failed by parsing the output only and not looking at the err and exitcode. Let me know if you think of any other way.

tests/projects/sync-env-alpine/wercker.yml
lake-of-dreams: This exists for all the alpine based images so far - do you mean we should just pin this to current latest - `alpine:3.6` ?

nigeldeakin:
>I have a concern that we're just trading in one issue for another if the user has multiple lines in their envvars (for example with ssh keys)
Yes, this doesn't solve the issue of multi-line env vars with alpine. But it does make single-line env vars work with alpine, so it reduces the scope of this issue.

However I'm concerned that if a multi-line env var is encountered this function will fail (and return an error) but without reporting anything to the user. The user might find that none of their env vars are being passed to the internal step, or that the after-steps are not run, and have no idea why. The only way to discover what has gone wrong will be to look for the warning messages in kibana. 

However if we logged a warning message to the step output this would mean that if the user has a multi-line env var then they will see this message even if they were not trying to pass the multi-line env var to the internal step. That might surprise existing users.










What happens if a multi-line env var is encountered on alpine? Is an error message displayed in the UI?

lake-of-dreams:
> > I have a concern that we're just trading in one issue for another if the user has multiple lines in their envvars (for example with ssh keys)
> > Yes, this doesn't solve the issue of multi-line env vars with alpine. But it does make single-line env vars work with alpine, so it reduces the scope of this issue.
> 
> However I'm concerned that if a multi-line env var is encountered this function will fail (and return an error) but without reporting anything to the user. The user might find that none of their env vars are being passed to the internal step, or that the after-steps are not run, and have no idea why. The only way to discover what has gone wrong will be to look for the warning messages in kibana.
> 
> However if we logged a warning message to the step output this would mean that if the user has a multi-line env var then they will see this message even if they were not trying to pass the multi-line env var to the internal step, or use after-steps. That might surprise existing users for whom the failure was not actually causing any problems.

For existing users - `SyncEnvironment` does not work at all in alpine - which would mean that indeed if they were dependent on a multi environment variable being passed from an step to one of these internal steps - their builds are still failing.  if a multi line environment variable is encountered now after this fix, it is simply ignored and will not be available - we `split` on newlines and then we check if it comes out as a `=` pair and then only export, and therefore the newline environment variable will still not be available

core/pipeline.go
lake-of-dreams: refactored error handling

core/pipeline.go
lake-of-dreams: fixed

core/pipeline.go
lake-of-dreams: made the changes

core/pipeline.go
lake-of-dreams: see https://github.com/wercker/wercker/pull/531#discussion_r229652961

lake-of-dreams:
As mentioned in the PR description, I do not think these changes are going to introduce any new issue - its fixing an existing issue of `SyncEnvironment` for builds using alpine partially. If there were any existing builds based on alpine and relying on `SyncEnvironment` for multi line environment variables like ssh keys, those would still be failing and therefore do not think any regression will be caused..

tests/projects/sync-env-alpine/wercker.yml
lake-of-dreams: If we just pin this to `alpine:3.6` - this test will not be useful for alpine images post `3.6`. Even if they switch the `env` library in a later release - the test will still continue to work as there is no change to existing logic of `env --null` - therefore I think it is more appropriate and useful to keep it pointing to latest

hatchan: We are not testing alpine is this test, we are testing if wercker is able to run on a image that contains a env that does not support `--null`. Thus we need to be sure that the image does contain this binary. 

If they do not supply something like `alpine:x.x.x`, then I guess `alpine:3.6` is fine, as changing this binary seems like a backwards incompatible change.

core/pipeline.go
```diff
@@ -311,3 +321,23 @@ func (p *BasePipeline) SyncEnvironment(sessionCtx context.Context, sess *Session
 func (p *BasePipeline) Docker() bool {
 	return p.config.Docker
 }
+
+func checkError(exit int, output []string, err error) error {
+	if exit != 0 {
```
hatchan: We first need to check if err is not nil. When a error is returned `exit` will be `-1`, if that happens than we loose the original `err`, which might contain something else besides the "Command exited with exit code" error.

core/pipeline.go
```diff
@@ -311,3 +321,23 @@ func (p *BasePipeline) SyncEnvironment(sessionCtx context.Context, sess *Session
...
+	}
+
+	if checkOutputEmpty(output) {
+		return fmt.Errorf("No output returned by env command from container")
+	}
+
+	return nil
+}
+
+func checkOutputEmpty(output []string) bool {
```
hatchan: If we're already putting all the checks on newlines above, then we might as well inline this as well.

core/pipeline.go
```diff
@@ -311,3 +321,23 @@ func (p *BasePipeline) SyncEnvironment(sessionCtx context.Context, sess *Session
...
+	return !(output != nil && len(output) > 0)
```
hatchan: This is more readable imo:

```suggestion
	return output == nil || len(output) == 0
```

core/pipeline.go
hatchan: I see, I guess we need to do the revert in a separate "SendChecked" command. As it looks like it only looks at the last command to determine the exit code.

tests/projects/sync-env-alpine/ENVIRONMENT
```diff
@@ -0,0 +1,4 @@
+X_REPO_OWNER=testowner
+X_REPO_NAME=testrepo
+X_REPO_FULL=$REPO_OWNER/$REPO_NAME
+X_REPO=$REPO_FULL
```
nigeldeakin: Can you add a multi-line environment variable here? This would improve the test by verifying that a multi-line environment variable does not prevent `SynEnvironment` from handling the other variables correctly.

nigeldeakin:
>  if a multi line environment variable is encountered now after this fix, it is simply ignored and will not be available 
Sounds good. I've suggested adding a test for this.

tests/projects/sync-env-alpine/wercker.yml
lake-of-dreams: ok. I have changed to `alpine:3.8` which is the current latest

core/pipeline.go
lake-of-dreams: Fixed

core/pipeline.go
lake-of-dreams: Fixed

core/pipeline.go
lake-of-dreams: Fixed

tests/projects/sync-env-alpine/ENVIRONMENT
lake-of-dreams: Added

lake-of-dreams: it seems it was not possible to add a variable containing a new line in ENVIRONMENT file due to the way we parse it in code. I now have added a SSH_KEY from a previous script step as a multi line var and it is ignored.

lake-of-dreams: External test for this  - https://dev.wercker.com/ravi-r-singhal/build-push-test/runs/build/5bdab5997dbfbf002545b6a4?step=5bdab5afed48c1000adb2357

nigeldeakin: Right. Or you could have added a prior step to set it.

nigeldeakin:
```
    - script:
        name: A script step
        code: |
          export long="This
          is
          a multi
          line
          step"
```

PR #538, opened 2018-11-21
Fix for WRKR-1429: RDD is deprovisioned before after-steps have been run

nigeldeakin:
This PR fixes https://jira.oraclecorp.com/jira/browse/WRKR-1429, which reported that when `docker:true` is specifed in a hosted application, the remote docker daemon is deprovisioned when the pipeline's "normal" steps have completed but before any after-steps have run. This means that unless the after-steps are very quick they will fail because there is no daemon to run them in.


nigeldeakin:
Failing test case: https://github.com/nigeldeakin/testcase2

See it fail in production at https://app.wercker.com/nigeldeakin/testcase2/runs/build/5bf59594567ae900076d60e4

See it work in staging, using my experiment:
https://dev.wercker.com/nigeldeakin/testcase2/runs/build/5bf5959497cdec0007fab811

nigeldeakin:
Thanks. Perhaps we should extract the main part of the pipeline into a separate function, and then use defer to invoke the buildFinisher at the end. Then we no longer need the explicit call.

PR #550, opened 2018-12-11
Wrkr 1476 - retry fetching step version and tarball for wercker-init step

lake-of-dreams:
https://jira.oraclecorp.com/jira/browse/WRKR-1476 - these changes add a retry when we see an error while either fetching the step version or downloading the step tarball - *only* for `wercker-init` step.

Tested locally by interrupting the network while fetching step version/downloading the tarball to verify that retry happens for `wercker-init` and behaviour for all other steps is left as it is. Tested using canary in dev.wercker to see there is no other impact


core/step.go
```diff
@@ -374,10 +380,15 @@ func (s *ExternalStep) Fetch() (string, error) {
 
 			stepInfo, err := client.GetStepVersion(s.Owner(), s.Name(), s.Version())
 			if err != nil {
-				if apiErr, ok := err.(*api.APIError); ok && apiErr.StatusCode == 404 {
-					return "", fmt.Errorf("The step \"%s\" was not found", s.ID())
+				if stepWerckerInitID != s.ID() {
```
hatchan: We need to do this for every step, not just wercker-init

core/step.go
```diff
@@ -374,10 +380,15 @@ func (s *ExternalStep) Fetch() (string, error) {
 
 			stepInfo, err := client.GetStepVersion(s.Owner(), s.Name(), s.Version())
```
hatchan: Add the retry to the `GetStepVersion` func and have it implicitly retry for 3 times, instead of doing it manually in the error handling if.

hatchan: Same applies to `GetTarball`

api/registry.go
```diff
@@ -68,5 +74,26 @@ func (r *WerckerStepRegistry) GetStepVersion(owner, name, version string) (*APIS
...
+}
+
+func (r *WerckerStepRegistry) getWithRetry(url string, authToken string) (*http.Response, error) {
+	var resp *http.Response
+	var err error
+	try := 0
+	maxTries := 3
+	for try < maxTries {
+		if try != 0 {
+			r.logger.Debugf("Retrying step url %s %d", url, try)
```
nigeldeakin: Will this get recorded in Kibana? If not then it should, I think.

lake-of-dreams: just checked that `debug` ones from cli does not make it to kibana, changed to `info` which does

nigeldeakin: Thanks

gilbode:
If this is for the tls handshake timeout you might also want to set that timeout higher.  Found an example here - https://github.com/mitchellh/goamz/issues/241

PR #553, opened 2019-01-04
Handle case where label spec is not of form label=value

nigeldeakin:
This PR fixes https://jira.oraclecorp.com/jira/browse/WRKR-1486

If you use the `internal/docker-push` step, and set the `labels` property to contain a label specification which is not of the form `label=value` then in hosted wercker the step fails without displaying any input or error.

If you do the same using the wercker CLI locally then it crashes with a panic.

This PR improves the handling of the `labels` property to catch this case and fail properly with a suitable error. 

A new unit test has been added to test this. 

In addition I have tested this in staging to confim that the error is displayed in the web UI. You can see this here:  https://dev.wercker.com/nigeldeakin/bedrock/runs/build/5c2f560a7a88ed000720fd83?step=5c2f56312ce5e50006cc1f4e


nigeldeakin:
According to https://jira.oraclecorp.com/jira/browse/WRKR-1486 the issue occurs when the `labels` property was set to `thisIsABadLabel`, which is invalid. 
```
    - internal/docker-push:
         username: $OCI_USERNAME
         password: $OCI_PASSWORD
         repository: iad.ocir.io/odx-pipelines/bihaber/test
         registry: https://iad.ocir.io/v2
         tag: $WERCKER_GIT_BRANCH-$WERCKER_GIT_COMMIT
         labels: thisIsABadLabel
```
This PR fixes that case as described above.

nigeldeakin:
I looked at https://github.com/wercker/kiddie-pool/blob/master/wercker.yml. This has
```
 labels: bundled-cli.branch=${BUNDLED_CLI_BRANCH} bundled-cli.commit=${BUNDLED_CLI_COMMIT}  
```
If this failed for you, what were the values of `${BUNDLED_CLI_BRANCH}` and `${BUNDLED_CLI_COMMIT} ` that caused the push to fail? If I set these to a valid branch name and a valid commit ID the push to OCIR works fine for me.

nigeldeakin:
Incidentally, it's fine for a label to be set to an empty string, or for the name of a label to be an empty string. So `label1=` is OK, `=value1` is OK, even `=` is OK.

PR #558, opened 2019-03-07
Make wercker-init to be an embedded step

gsserge:


core/wercker_init_step.go
```diff
@@ -0,0 +1,103 @@
...
+#               be working.
+alias sudo=""
+
+# Make sure we fail on all errors
+set -e
+`
+
+// NewWerckerInitStep is a special step.
+func NewWerckerInitStep(options *PipelineOptions) (*WerckerInitStep, error) {
+	werckerInit := "wercker/wercker-init@2.0.0"
```
hatchan: Maybe change the version to distinguish it from the registry.

PR #559, opened 2019-03-07
Fix interrupts

nigeldeakin:
This PR fixes https://jira.oraclecorp.com/jira/browse/WRKR-1565 (When a run is aborted, the step is still shown as running)

**Changes to `SetupEnvironment` in `cmd/runner.go`**

I have changed the working of the two signal handlers (for INT and TERM signals) used in this function.

**`boxCleanupHandler`**

Before this PR this function created a single signal handler just before the end and leaves it in place for the rest of the run. When signalled, this shuts down the box and calls `os.Exit()`. 

This is the only handler that actually shuts down the program, so I have moved it to the start of the function. Without this change, an interrupt during the "setup environment" step does not terminate the pipeline.

I have renamed the variable to `boxCleanupHandler`

**`stepInterruptedHandler`**

In addition to this, I have added a second signal handler just before the call to `StartStep` (which sends the "step started" event). This means that if an interrupt occurs between that point and the end of the function the "step aborted" event will sent. 

(Note that the code for this handler is the same as in `RunStep` below.)

**Changes to `RunStep` in `cmd/runner.go`**

This function is called to run each normal step.

Before this PR, this function registered a signal handler which sends the "step aborted" event in the event of an interrupt.  I have made some change to this.

* The signal handler is now registered BEFORE the call to `StartStep` (which sends the "step started" event) rather than AFTER. This avoids having a small window after the "step started" event has been sent when the handler has not yet been registered. If an interrupt had occurred during that interval the step would have been left in the "running" state.

* The signal handler now explicitly sets the "step aborted" message to "Step interrupted".  Without this change there might not be a message at all, or a message such as "context cancelled".

* If the signal handler is fired then it creates a new `StepResult` rather than use the existing one, This avoids any possibility that it will be overwritten if the step comes to an end whilst it is running.

**Changes to `TestRunnerStepFailedOnInitEnvError` in `cmd/runner_test.go`**

This test uses a mock Step object. I have changed this to set the ID, Name and DisplayName fields, which were previously unset. 

This change is necessary because I have added logging which expects these to be non-nil


util/signal.go
```diff
@@ -58,6 +58,7 @@ func (s *SignalMonkey) Remove(fn *SignalHandler) {
 			copy(s.handlers[i:], s.handlers[i+1:])
 			s.handlers[len(s.handlers)-1] = nil // or the zero value of T
 			s.handlers = s.handlers[:len(s.handlers)-1]
+			return
```
hatchan: There's no guarantee that ID is unique in the collection, so shouldn't we go through the entire array (instead of stopping early)?

nigeldeakin: I added this change because if this code removes an element from the collection it is iterating over and then continues to iterate, `x` will be nil which means that `x.ID1 causes a panic.

nigeldeakin: I have now pushed a fix to `SignalMonkey.Remove` which can cope with non-LIFO remove order without assuming that IDs are unique.

PR #560, opened 2019-03-08
Turn on enable-dev-steps flag by default

gsserge:


gsserge:
Because currently `--enable-dev-steps` needs to be supplied to enable local steps (downloaded from the store and specified by the file path in the yml file).

PR #562, opened 2019-03-13
Add support for envFiles in workflows

hatchan:
Also fixes various minor documentation and usability issues


hatchan:
> Looks good to me. Question on the XXX_ variables, how are they supposed to work?

Any envvar defined on app.wercker.com as protected will get the `XXX_` prefix. This means that they won't show up in logs etc.

util/environment.go
```diff
@@ -34,6 +34,15 @@ type Environment struct {
 	Order  []string
 }
 
+// DefaultEnvironment uses the default strategy of using the current
```
nigeldeakin: This comment is incomplete (the current what?)

util/environment.go
```diff
@@ -34,6 +34,15 @@ type Environment struct {
...
+func DefaultEnvironment(envfile string) *Environment {
+	env := NewEnvironment(os.Environ()...)
+	env.LoadFile(envfile)
+	env.PassThruProxyConfig()
```
nigeldeakin: I realise this is not new, but I'm surprised that this code doesn't inform the user if the file they specified does not exist.

util/environment.go
```diff
@@ -75,7 +84,7 @@ func (e *Environment) AddIfMissing(key, value string) {
 	}
 }
 
-// Add proxy configuration as public
+// PassThruProxyConfig adds proxy configuration as public environment variables.
 func (e *Environment) PassThruProxyConfig() {
```
nigeldeakin: My suggestion is
"PassThruProxyConfig ensures that environment variables that configure a proxy (such as HTTP_PROXY) are passed from the user's environment to the pipeline

util/environment.go
```diff
@@ -141,14 +150,14 @@ var proxyEnv = [...]string{
 	"NO_PROXY",
 }
 
-// Collect passthru variables from the project
+// GetPassthru gets the passthru variables from the project
```
nigeldeakin: This function is getting these variables from the user's environment, not the project, I think.

util/environment.go
```diff
@@ -141,14 +150,14 @@ var proxyEnv = [...]string{
...
 func (e *Environment) GetPassthru() (env *Environment) {
-	return e.passthru(public)
+	return e.passthru(publicPrefix)
 }
 
-// Collect the hidden passthru variables
+// GetHiddenPassthru gets the hidden passthru variables
```
nigeldeakin: This function is getting these variables from the user's environment, not the project, I think.

util/environment.go
nigeldeakin: Pass change "passthru variables" to "passthru variables (those starting with publicPrefix)". Otherwise you have to read the code to know what a hidden passthru variable is.

util/environment.go
nigeldeakin: Pass change "hidden passthru variables" to "hidden passthru variables (those starting with privatePrefix)". Otherwise you have to read the code to know what a hidden passthru variable is.

cmd/workflow.go
```diff
@@ -136,6 +136,10 @@ func getPipelineOptions(opts *core.WorkflowOptions, pipeline *core.WorkflowPipel
 	po.RunID = bson.NewObjectId().Hex()
 	po.ShouldArtifacts = true
 
+	if pipeline.EnvFile != "" {
+		po.HostEnv = util.DefaultEnvironment(pipeline.EnvFile)
```
nigeldeakin: This is the actual change. Looks good to me.

util/environment.go
hatchan: That is the current behaviour.

hatchan: The default flag for environment is always set (`ENVIRONMENT`), most likely it will not exists, so would result in all kinds of warnings.

util/environment.go
hatchan: Changed it to the following: 
```
// PassThruProxyConfig prefixes any proxy environment variables defined on the
// host with publicPrefix. So they will be available during a build.
```

util/environment.go
hatchan: changed it to:
```
// GetPassthru gets the environment variables from e which should be exposed to
// the build as normal environment variables.
```

util/environment.go
hatchan: Changed it to:
```
// GetHiddenPassthru gets the environment variables from e which should be
// exposed to the build as hidden environment variables.
```

hatchan:
Went through the text of the documentation.

hatchan:
Copyright changes will come in the next commit.

cmd/main.go
```diff
@@ -463,7 +455,8 @@ var (
 
 			opts.WorkflowName = c.Args().Get(0)
 			if opts.WorkflowName == "" {
-				cliLogger.Errorln("Missing workflow name to run")
+				cliLogger.Errorln("Missing workflow name to run\n")
+				cli.ShowCommandHelp(c, "workflow")
 				os.Exit(1)
 			}
```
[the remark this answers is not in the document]
hatchan: See https://github.com/wercker/wercker/pull/562/commits/87e564833257fe5d56f5b7cf7530605391c604eb

util/environment.go
hatchan: Text has been updated

PR #566, opened 2019-03-29
Allow wercker CLI to be run locally with a web proxy set

nigeldeakin:
This PR fixes https://jira.oraclecorp.com/jira/browse/WRKR-1569, which reported that if the wercker CLI was run locally, and `http_proxy` was set to configure the use of a web proxy, then the pipeline could not connect either to service containers or to containers started using `internal/docker-run`. This was because connections to these containers were routed to the proxy, which knew nothing about them.

This PR fixes this by adding the names of services and containers started using `internal/docker-run` to the `no_proxy` and `No_PROXY` environment variables. 

`no_proxy` is only changed if `http_proxy` or `https_proxy` are set. 
`NO_PROXY` is only changed if `HTTP_PROXY` or `HTTPS_PROXY` are set.

For services this can be done right at the start of the pipeline, before any environment variables are set. 

For containers started by `internal/docker-run` this must be done by the step itself. This means that it has to obtain the prior values of environment variables, and set new values for these variables, by sending commands to the pipeline container. 

To make sure that changes made by the `internal/docker-run` step are seen in after-steps I have added a new call to `SyncEnvironment` at the end of the main part of the pipeline, to ensure that any changes are captured and propagated to the container used for the after-steps. This change has the happy side-effect of fixing https://jira.oraclecorp.com/jira/browse/WRKR-1082  (that Env vars set in pipeline steps are not propagated to after-steps). 

This is tested using a new integration test. Some tests use alpine, some use ubuntu, to help detect any issues that might arise.


nigeldeakin:
I have pushed a further commit which fixes a bug in `SendChecked`, which sends a command to the container and returns the output. 

`SendChecked` is used in many places but most are not interested in the output that is returned. However `SyncEnvironment` is very dependent on receiving the correct output, and this PR (and its test) is reliant on `SyncEnvironment` working correctly. That is why this PR has uncovered this existing bug, which causes the new test to fail when run on Mac.

The `SendChecked` sends a command to the container followed by a command which echoes a sentinel string. It receives the output from the container, and assumes that when it receives the sentinel string back again it has received all the output.

This does not always work if the command writes to standard error, because the sentinel (which is written to standard output) may be received before all of standard error is received.

This is a problem because when `SyncEnvironment` uses `SendChecked` it deliberately sends a command (`env --null`) which it expects to fail on alpine, and searches the output for an error message which will have been sent to standard error. Because the sentinel may "overtake" the output from standard error  `SendChecked` sometimes returns only the first part of the error message, with the remainder being returned at the start of the next call to  `SendChecked`. This behaviour breaks  `SyncEnvironment`, which breaks this new test case.

The fix is to change `SendChecked` to send out *two* sentinels, which echo on standard output and standard error. Only when *both* have been received back can we assume that all the output has been received. 

This fix makes a dramatic difference when this test is run on Mac. Without it the test fails more than half the time. With it it never fails.

PR #569, opened 2019-05-23
Allow using escaped '$' in env var values

desagar:
Allow users to use a literal '$' sign in environment variables, escaping it as $$

Currently the only way the '$' sign is treated in env vars is to interpolate the value of another env var. There is no way to use a literal '$' since escaping it with a backslash has no effect.

In this PR, I'm proposing to use '$$' as the escape sequence for a single '$' sign. This will need to be documented.
The preferred escape sequence is '\\$' but that may not be backward compatible.


desagar:
Closing because this is more involved than what this PR tries to do.

